PR #40, opened 2015-01-06
ENH: Allow silent operation

Eric89GXL:
I don't want my coverage reports to print on my CI targets, and using `--cov-report=html | xml` adds time overhead. This adds an option to suppress printing altogether. Let me know if this seems useful!

### DIFF
--- a/pytest-cov/README.rst
+++ b/pytest-cov/README.rst
@@ -190,14 +190,20 @@ The terminal report with line numbers::
     TOTAL                  353     20    94%
 
 
-The remaining three reports output to files without showing anything on the terminal (useful for
-when the output is going to a continuous integration server)::
+These three report options output to files without showing anything on the terminal::
 
     py.test --cov-report html
             --cov-report xml
             --cov-report annotate
             --cov myproj tests/
 
+The final report option can also suppress printing to the terminal::
+
+    py.test --cov-report= tests/
+
+This mode can be especially useful on continuous integration servers, where a coverage file
+coverage file is needed for subsequent processing, but no local report needs to be viewed.
+For example, test run on Travis-CI could produce a .coverage file for use with Coveralls.
 
 Coverage Data File
 ------------------
--- a/pytest-cov/README.rst
+++ b/pytest-cov/README.rst
@@ -68,7 +68,7 @@ subprocesses.
 
 Running centralised testing::
 
-    py.test --cov myproj tests/
+    py.test --cov-source myproj tests/
 
 Shows a terminal report::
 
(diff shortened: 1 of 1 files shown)


Eric89GXL:
BTW if I don't have this option enabled, on my project after my coverage report prints I get the following error (on the `2.0` branch):
```
  File "<string>", line 9, in <module>
  File "/usr/lib/python2.7/dist-packages/_pytest/config.py", line 37, in main
    return config.hook.pytest_cmdline_main(config=config)
...
  File "/usr/lib/python2.7/dist-packages/_pytest/core.py", line 314, in execute
    res = method(**kwargs)
  File "/home/larsoner/.local/lib/python2.7/site-packages/pytest_cov-1.8.1-py2.7.egg/pytest_cov.py", line 163, in pytest_terminal_summary
    assert total is not None, 'Test coverage should never be `None`'
```
Should I open an issue for this, or is it a known issue?

Eric89GXL:
(This problem does not exist on the `pip` version because the offending `assert` line is not in the 1.8.1 source.)

Eric89GXL:
Ahh, ignore the bug report, I realized my problem is that I didn't install the updated `cov_core` in addition to the updated `pytest-cov`.

In any case, let me know what you think of the PR.

schlamar:
You can disable output with `--cov-report=` ([see](https://github.com/schlamar/pytest-cov/blob/2.0/pytest-cov/test_pytest_cov.py#L574))

(Note to myself: update README)

Eric89GXL:
Whoops, sorry for the noise! I could have sworn I tried that, because I saw a condtional that seemed to check for that in the code -- I just tested it and it does indeed work.

I've updated the PR (w/squashed commits) to update the README.rst, feel free to merge, comment, or just close the PR if you'd prefer to update it yourself.

pytest-cov/README.rst
```diff
@@ -190,14 +190,20 @@ The terminal report with line numbers::
...
+These three report options output to files without showing anything on the terminal::
 
     py.test --cov-report html
             --cov-report xml
             --cov-report annotate
             --cov myproj tests/
 
+The final report option can also suppress printing to the terminal::
+
+    py.test --cov-report= tests/
```
schlamar: This is no valid command line, please add `--cov-source myproj`. This is the recommended invocation from 2.0 on, `--cov myproj` is deprecated. And while on it, please update it in the example above :)

pytest-cov/README.rst
```diff
@@ -190,14 +190,20 @@ The terminal report with line numbers::
...
+
+This mode can be especially useful on continuous integration servers, where a coverage file
+coverage file is needed for subsequent processing, but no local report needs to be viewed.
```
schlamar: `coverage file` is doubled ;)

schlamar:
Had some minor remarks, but otherwise LGTM.

Eric89GXL:
Thanks for the quick review, I updated all instances of `--cov myproj` to `--cov-source myproj`.

schlamar:
Thanks :cake:

PR #41, opened 2015-01-12
Uniformize & Vectorize badges

DavidJFelix:


schlamar:
Was included in #58 but thanks anyway :)

PR #45, opened 2015-02-12
multiprocessing_start finalizer fix

aldanor:
This fixes problems like this (when running `py.test` on a bunch of parallelized tests):
```python
Traceback (most recent call last)                               :
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
  File "env/lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
```

### DIFF
--- a/cov-core/cov_core.py
+++ b/cov-core/cov_core.py
@@ -10,9 +10,10 @@ import os
 
 def multiprocessing_start(obj):
     cov = cov_core_init.init()
-    import multiprocessing.util
-    multiprocessing.util.Finalize(
-        None, multiprocessing_finish, args=(cov,), exitpriority=1000)
+    if cov is not None:
+        import multiprocessing.util
+        multiprocessing.util.Finalize(
+            None, multiprocessing_finish, args=(cov,), exitpriority=1000)
 
 
 def multiprocessing_finish(cov):


aldanor:
Ok, it turns out that's probably not the right way to do it according to tests... Hmm...

aldanor:
To explain where `cov` can be `None` -- in `cov_core_init.init()`, if `cov_config` is `None` (which it can be in a child process, I guess), then it doesn't return anything, so `cov` is `None` which causes further problems with finalizers.

schlamar:
I assume Travis messed somehow up, in general your change looks good.

Please rebase against my latest commit then tests should pass.

schlamar:
Fixed in #49. Thanks anyway! :)

PR #49, opened 2015-03-27
Remove MP Finalize if cov didnt initialize.

alisaifee:
There may be cases where cov_core_init.init() returns None. 
Currently, multiprocessing_finish raises an AttributeError.
```
Traceback (most recent call last):
  File "/..../lib/python2.7/multiprocessing/util.py", line 274, in _run_finalizers
    finalizer()
  File "/.../lib/python2.7/multiprocessing/util.py", line 207, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/.../lib/python2.7/site-packages/cov_core.py", line 19, in multiprocessing_finish
    cov.stop()
AttributeError: 'NoneType' object has no attribute 'stop'
```

It's a bit hard to reproduce with a unit test but i'm seeing the error when I run py.test with the `pytest-cov` plugin installed but not invoked

PR #52, opened 2015-05-11
Protect multiprocessing_finish against None coverage object.

fmichea:
Hi,

The `cov` parameter of the `multiprocessing_finish` hook is given by the `multiprocessing_start` function that calls `cov_core_init.init()`. This init function can return a `None` value if it fails with an exception (ignores it and moves on) or the environment is not configured properly (doesn't create a coverage object).

This protects the hook from calling the `stop` and `save` function on a `None` value. Another part of the fix might be to warn when the coverage environment configuration is not set, instead of ignoring.

### DIFF
--- a/cov-core/cov_core.py
+++ b/cov-core/cov_core.py
@@ -16,8 +16,9 @@ def multiprocessing_start(obj):
 
 
 def multiprocessing_finish(cov):
-    cov.stop()
-    cov.save()
+    if cov is not None:
+        cov.stop()
+        cov.save()
 
 
 try:


schlamar:
Thanks but this should already be handled by #49?!

fmichea:
Oh you're right, sorry about that closing this pr. Thanks for addressing this!

fmichea:
Actually Marc, I checked and asked some of my friends to check too, and none of us can understand where this commit is on your branches. If we look at your commit history, or at the HEAD of your branch when cloning, we cannot find this change.

It is not the only one, I cannot find #50 either for instance. How do you merge them? Do you keep those some other place? Could it be possible that you merged them and then force pushed some of your local changes to github by any chance? That would have overwritten the changes.

Thanks for your help understanding what is happening here!

schlamar:
I assume this is a bug on GitHub's side. I already notified them. I had the commits from #50 still locally so I just pushed them.

schlamar:
Thanks for reporting this BTW :cake:

jgehrcke:
Can we push a new release to PyPI for fixing this? Currently it affects everybody installing pytest-cov:
```
$  pip install pytest-cov
[...]
Successfully installed cov-core-1.15.0 pytest-cov-1.8.1
```
test_mp.py:
```python
import multiprocessing

def testmp():
    p = multiprocessing.Process(target=lambda: None)
    p.start()
    p.join()
```

and execution:

```
 $ py.test -s test_mp.py
================================================= test session starts =================================================
platform linux -- Python 3.4.3 -- py-1.4.30 -- pytest-2.7.2
rootdir: ***, inifile:
plugins: cov
collected 1 items

test_mp.py Traceback (most recent call last):
  File "/***/3.4.3/lib/python3.4/multiprocessing/util.py", line 252, in _run_finalizers
    finalizer()
  File "/home/jang/nobackup/pyenv/versions/3.4.3/lib/python3.4/multiprocessing/util.py", line 185, in __call__
    res = self._callback(*self._args, **self._kwargs)
  File "/***/python3.4/site-packages/cov_core.py", line 19, in multiprocessing_finish
    cov.stop()
AttributeError: 'NoneType' object has no attribute 'stop'
.

============================================== 1 passed in 0.02 seconds ===========================================
```
Thanks!

schlamar:
Released!

jgehrcke:
Nice, thanks!

PR #55, opened 2015-06-28
Don't bother adding finalizer if no coverage object was created

maguro:
There's no need to stop and save the coverage object if one wasn't created.


schlamar:
Thanks but this was already be handled by #49. This PR was somehow stripped from the history but I just re-added it.

PR #57, opened 2015-06-29
CMD API, --cov-source activate the pluggin even without --cov

Natim:


schlamar:
Thanks, but you need to handle this [here](https://github.com/Natim/pytest-cov/blob/487c287597563dddf524bbc62483da5331c3919c/pytest-cov/pytest_cov.py#L48), too.

And your test case is broken (wrong name on checking coverage results).

Natim:
I fallback to use pytest-cover instead.

PR #28, opened 2014-11-22
[WIP] 2.0

schlamar:
Todo: tracked in #58 from now on.

- [x] import cov-core into repository
- [x] run tests against cov-core master, too
- [x] PEP8, run flake8 against all projects
- [x] use `data_suffix=True` instead of building it manually
- [x] allow to specify data_file (https://github.com/schlamar/cov-core/issues/4) 
- [x] ~~better way to get coverage data from xdist slaves (e.g. transmit content of coverage data file)~~ there is probably no better way
- [x] test COVERAGE_FILE environment variable
- [x] allow source from coveragerc (https://github.com/schlamar/pytest-cov/pull/21)
- [x] fail under (#31)
- [x] test COVERAGE_FILE env variable
- [x] CHANGELOG(s)
- [x] update classifiers in setup.py

Ideas:

- ~~use coverage.py's way of measuring subprocesses~~ (http://nedbatchelder.com/code/coverage/subprocess.html)
  single item which will break backwards compatibility, so no.

PR #58, opened 2015-06-30
Merge cov-core and pytest-cov packages and other fixes

ionelmc:
First, I'm terribly sorry to make a PR like this, I got a bit impatient and there wasn't any activity on `pytest-cov` for a while so I made a [fork](https://github.com/ionelmc/pytest-cover). This PR includes the changes from that fork. Please don't send me back to the loony bin :grin: 

Changes I did and some of the goals behind them:

* Keep the same arguments as for `pytest-cov 1.x`. No breaking changes, but still bring in the fixes regarding `.coveragerc` handling:
  
  * Using bare `--cov` will use the sources specified in `.coveragerc`. Handling [here](https://github.com/ionelmc/pytest-cov/blob/single-package-and-other-fixes/src/pytest_cov/plugin.py#L43-L46). This means the `--cov-source` option from the 2.0 branch is gone.
  * Using `--cov-report=term` will report missing lines if `.coveragerc` has `show_missing = True`
  * The option `--cov-fail-under` is automatically activated if `.coveragerc` has `fail_under = ...` (only if using coverage 4.0)
  * The option `--cov-min` is renamed to `--cov-fail-under` to keep consistency with coverage docs (so that users will have an easier time finding the option).

* Fix the `.pth` activation. I found a solution that works when installing from install, easy_install, wheels and develop. Uninstall also works, but only for wheels.

* Some improvements for CI (separate testing with xdist and windows CI via AppVeyor)

* Merging the `cov-core` and `pytest-cov` packages to simplify things like the release process or compatibility problems caused by mismatched versions. I took an opinionated turn here I know but there's value in focusing on a single test tool.

## Todo

- [x] classifiers
- [x] update setup.py (remove "Forked from `pytest-cov`", maybe other fixes)
- [x] update LICENSE (re-add me as author, you can leave yours)
- [x] update changelog
- [x] update Acknowledgements in README
- [ ] update release date (don't know it yet)
- [ ] sort out command line options


schlamar:
> Using bare --cov will use the sources specified in .coveragerc. Handling here. This means the --cov-source option from the 2.0 branch is gone.

The reason for deprecating `--cov` with parameters is that per convention any command line parameter should have either arguments or no argument at all. Another option would be to keep the behavior as it is and introduce a new parameter - e.g. `--covrc` - which activates coverage and reads sources from `.coveragerc`. 

Or another option would be to always activate the plugin if there is a `.coveragerc` file and provide an option to disable it, something like `--cov-disable`.

schlamar:
Thanks for contributing, BTW. Looks very great in general :cake:

schlamar:
Is there anything left from the todo list in https://github.com/schlamar/pytest-cov/pull/28 (minus the points which are obviously obsolete).

ionelmc:
> The reason for deprecating --cov with parameters is that per convention any command line parameter should have either arguments or no argument at all. Another option would be to keep the behavior as it is and introduce a new parameter - e.g. --covrc - which activates coverage and reads sources from .coveragerc.

> Or another option would be to always activate the plugin if there is a .coveragerc file and provide an option to disable it, something like --cov-disable.

I agree it's not the cleanest cmdline api. Had `pytest-cov` not been already released with this `--cov` bi-valence I'd agree 100% with you. However, it's already released - are we getting any serious advantages by making users change their commands?

ionelmc:
> Is there anything left from the todo list in #28 (minus the points which are obviously obsolete).

I think this it's missing:

* [ ] update classifiers in setup.py

The rest should be fixed.

We might need to flatten out a bit the current changelog.

schlamar:
> Had pytest-cov not been already released with this --cov bi-valence I'd agree 100% with you.

It's not released yet, only pending in the 2.0 branch ;)

schlamar:
> update classifiers in setup.py

I guess we can use the classifiers from https://github.com/ionelmc/pytest-cover/blob/master/setup.py ?

ionelmc:
> It's not released yet, only pending in the 2.0 branch ;)

You are right. But what if we just add a `--cov-enable` option? (to avoid deprecating `--cov=foobar`). Hmmm, that's not pretty either ...

Do people want the `--cov-source`?

ionelmc:
> I guess we can use the classifiers from https://github.com/ionelmc/pytest-cover/blob/master/setup.py ?

Then it's fixed :)

schlamar:
> Or another option would be to always activate the plugin if there is a .coveragerc file and provide an option to disable it, something like --cov-disable.

I guess this is my favorite right now. Any thoughts or other ideas?

ionelmc:
> I guess this is my favorite right now. Any thoughts or other ideas?

I suspect explicit opt-out would annoy some people.

What disadvantages do we have with the dual `--cov` / `--cov=path` arguments?

schlamar:
> What disadvantages do we have with the dual --cov / --cov=path arguments?

I'll still think about it. :) But I guess the alternatives are not really better so we can leave it like this.

I've added some minor points as todo list to the first post. But in general it looks really great. I think we are almost good to go and can release this soon :)

ionelmc:
Alright, I've fixed those and squashed a little bit the unforking changes.

schlamar:
One last small thing: Rename `tests/test_pytest_cover.py` back :) 

Then I'll merge everything back in master and have a last look before release.

Thank you for this great PR :cake:

schlamar:
I'll do this on my own, merging it now.

ionelmc:
Ah yes, I'm a bit slow today.

About releasing, this is what I usually do:
```
rm -rf dist build src/*.egg-info; python3.4 setup.py clean --all sdist && twine upload dist/*
```

PR #59, opened 2015-07-06
Avoid warning about missing coverage data

ionelmc:
(just like ``coverage.control.process_startup``).

Now that the `.pth` install works so well we can get these benign "No data was collected" warnings all over the plate. They are quite annoying if you do assertions on subprocess output.


ionelmc:
This is odd. I don't see any build status ...

schlamar:
> This is odd. I don't see any build status ...

Works for me. Thanks :)

Issue #60, opened 2015-07-07
2.0 Release


schlamar:
- [x] remove usage and reference from docs
- [x] document release process
- [x] #69

@ionelmc

ionelmc:
Alright, gonna do this later today I hope

ionelmc:
Yey! Release time! :)

schlamar:
Released and added you as maintainer on PyPI! Thanks for your help!! :cake:

PR #61, opened 2015-07-08
Remove placeholder pages and add a release process.

ionelmc:
Ref #60.

Post-release process intentionally left out. I need to check how easy is to "dev-bump" with bumpversion (eg: v1.2.3 to v1.2.4dev)


schlamar:
Thanks :)

Issue #54, opened 2015-06-19
pytest-cov==1.8.1 + pytest-xdist does not show coverage report 


diefans:
the latest `pytest-cov==1.8.1` together with xdist does not show the coverage report anymore.

`pytest-cov==1.6` works

```
# pip freeze | grep pytest
pytest==2.7.1
pytest-cache==1.0
pytest-cov==1.8.1
pytest-pep8==1.0.6
pytest-random==0.2
pytest-xdist==1.12
```

ionelmc:
@diefans Is the report empty? Do you get any warnings?

diefans:
It is just not there. No warnings. As I would have called it without `--cov package`

ionelmc:
Does it still reproduce with the master branch (`pip install https://github.com/schlamar/pytest-cov/archive/master.zip`)?

diefans:
Master does not fix it :(

ionelmc:
Show the configurations you use (setup.cfg, pytest.ini, .coveragerc etc).

Also, show a tree of your project (run `tree`).

diefans:
my pytest section looks like:
```
[pytest]
addopts=-rf --strict --pep8 --cov-report term-missing --cov-config coveragerc --color yes --random --random-group
pep8maxlinelength = 120
pep8ignore=
    */migrations/* ALL
```
`.coveragerc`
```
# .coveragerc to control coverage.py
[run]
branch = True

[report]
# Regexes for lines to exclude from consideration
exclude_lines =
    # Have to re-enable the standard pragma
    pragma: no cover

    # Don't complain about missing debug-only code:
    def __repr__
    if self\.debug

    # Don't complain if tests don't hit defensive assertion code:
    raise AssertionError
    raise NotImplementedError

    # Don't complain if non-runnable code isn't run:
    if 0:
    if __name__ == .__main__.:

ignore_errors = True
precision = 2
omit=
```


pytest and plugins
```
# pip freeze | grep pytest
pytest==2.7.2
pytest-cache==1.0
pytest-cov==2.0.0
pytest-ipdb==0.1.dev2
pytest-pep8==1.0.6
pytest-random==0.2
pytest-xdist==1.12
```

ionelmc:
Project layout? (run `tree`)

diefans:
Its a bit big but you can also checkout http://github.com/diefans/python-arangodb
```
.
├── build
│   ├── bdist.linux-x86_64
[...]

PR #65, opened 2015-07-14
fix coverage report with `looponfail` (#54)

schlamar:
fixes regressions from 5d9ba4b and a123ac1

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -59,6 +59,9 @@ def pytest_configure(config):
     """Activate coverage plugin if appropriate."""
     if config.getvalue('cov_source'):
         if not config.pluginmanager.hasplugin('_cov'):
+            if not config.option.cov_report:
+                config.option.cov_report = ['term']
+
             plugin = CovPlugin(config.option, config.pluginmanager,
                                start=False)
             config.pluginmanager.register(plugin, '_cov')
@@ -85,6 +88,7 @@ class CovPlugin(object):
         self.cov = None
         self.cov_controller = None
         self.failed = False
+        self._started = False
         self.options = options
 
         is_dist = (getattr(options, 'numprocesses', False) or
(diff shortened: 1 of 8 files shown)


schlamar:
@diefans this should fix #54. Can you please confirm this?

Any idea how to write a test case for this? As already said, `looponfail` did never quit in my tests...

cc @ionelmc

diefans:
I can confirm this fix

ionelmc:
Isn't looponfail supposed to exit when all tests pass? Nevertheless, tests shouldn't get stuck - I can write a test for this after I fix that other issue.

schlamar:
> Isn't looponfail supposed to exit when all tests pass?

Didn't work like this for me...

diefans:
> Isn't looponfail supposed to exit when all tests pass?

no, it will run forever. It starts a new test run if a watched file is touched. So you can pretty well run your tdd session while watching your tests. If tests fail, only those which have failed are run again and after fixing those a complete test run is performed.

src/pytest_cov/plugin.py
```diff
@@ -59,6 +59,9 @@ def pytest_configure(config):
     """Activate coverage plugin if appropriate."""
     if config.getvalue('cov_source'):
         if not config.pluginmanager.hasplugin('_cov'):
+            if not config.option.cov_report:
```
ionelmc: Is this check really necessary? I'm curious in what situation this would be hit.

schlamar: Yes, because `pytest_load_initial_conftests` is not run in the xdist process and then coverage would be collected but not reported.

ionelmc: I tried to remove those lines and it seemed to work fine without 'em. Did I miss something?

schlamar: Yes ;) If you comment out this line https://github.com/schlamar/pytest-cov/blob/8209d9dbfdc23f51b58a9964798f2f82d79e685c/tests/test_pytest_cov.py#L516 it has an effect ;)

PR #67, opened 2015-07-14
Add a test for the fix in #65.

ionelmc:


schlamar:
Thanks, looks good!

schlamar:
@ionelmc any idea why this fails now on Pypy: https://travis-ci.org/schlamar/pytest-cov/builds/72790010

PR #76, opened 2015-08-01
Use https for all badges

moreati:
Fixes an insecure content warning when viewing the package on PyPI e.g. https://pypi.python.org/pypi/pytest-cov


ionelmc:
Good catch. Thanks!

PR #80, opened 2015-08-13
add option to append coverage instead of deleting the .coverage file …

### LINKED ISSUE #79: append coverage over several runs
I am migrating test to py.test + coverage.

previously we ran the tests with `coverage run -a manage.py test sms`

to append the coverage of all tests. pytest-cov does always erase the .coverage file.

I'd like an option --cov-append that will combine the coverage of all test results

cleder:
…fixes #79

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -11,3 +11,4 @@ Authors
 * Federico Ceratto - http://firelet.net/
 * Josh Kalderimis - http://blog.cookiestack.com/
 * Ionel Cristian Mărieș - http://blog.ionelmc.ro
+* Christian Ledermann - https://github.com/cleder
--- a/README.rst
+++ b/README.rst
@@ -255,7 +255,9 @@ tests run on Travis-CI could produce a .coverage file for use with Coveralls.
 Coverage Data File
 ==================
 
-The data file is erased at the beginning of testing to ensure clean data for each test run.
+The data file is erased at the beginning of testing to ensure clean data for each test run. If you
+need to combine the overage of several test runs you can use the `--cov-append` option to append
+this coverage data to coverage data from previous test runs.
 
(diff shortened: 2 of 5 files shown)


ionelmc:
Hey. Thanks for making the change :)

But can you rename the option to `--cov-append`? All the other options start with `--cov`.

schlamar:
We should have a test for this.

cleder:
added tests and some documentation to README

ionelmc:
I was wondering if we'd be better of with assertions on stdout (like here https://github.com/pytest-dev/pytest-cov/blob/master/tests/test_pytest_cov.py#L230-L235). That would ensure the loading works properly (your current assertions with the size might pass if you remove ``--cov-append``). What do you think?

ionelmc:
Alright, this is included in master and fixed up a bit for xdist support. Thanks.

aldanor:
Wonder if the coverage is persisted between runs? (in which case, how would it be possible to reset it so it still gets combined for all environments but not across different runs?)

ionelmc:
@aldanor I think you are looking for something like https://github.com/ionelmc/python-nameless/blob/master/tox.ini#L118-L122

aldanor:
@ionelmc I see, thanks. Feels a bit like a hack of course, because if you pass environments explicitly, e.g. `-e py27 -e py34`, this wouldn't work..

I guess the best use case would be gathering combined coverage for all Python versions within one run -- isn't it quite common?

ionelmc:
@aldanor pytest don't support multiple python version testing, so we can't do anything special regarding that in pytest-cov. It's something in Tox's domain - it it has some inconveniences as you saw. If you want to use detox see [this](https://hynek.me/articles/testing-packaging/).

The-Compiler:
One could probably write a tox plugin to do that though.

PR #83, opened 2015-08-20
Coverage 4.0b2 support

ionelmc:
These changes need a version later than 4.0b1 (the b1 release has this issue https://bitbucket.org/ned/coveragepy/issues/399/coverageexception-cant-combine-line-data). cc @The-Compiler

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,0 +1,4 @@
+[bumpversion]
+current_version = 2.0.0
+commit = True
+tag = True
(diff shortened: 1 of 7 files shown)


The-Compiler:
It already worked for me before (because I don't actually launch subprocesses which generate coverage data - I launch them to test subprocess handling :wink:), but it still works for me with the PR FWIW.

Thanks, @ionelmc!

As for 8ddea05, I wonder why those APIs were made private, and if pytest-cov really needs them - using them even though they're private is asking for trouble, isn't it? /cc @nedbat

PR #95, opened 2015-10-01
Fix handling when tests change working directory.

### LINKED ISSUE #77: pytest-cov makes test fail by putting a temporary file into cwd
i have tests that check if some specific files have been extracted into cwd.
they fail when using pytest-cov because some .coverage* file gets created there.

is there any way to avoid this?

### LINKED ISSUE #94: Calling subprocess fails (and outright crashes on python 2.6) when given a relative path for --cov-config
As I documented over in astropy/astropy-helpers#193, if I specify a relative path to a coveragerc file with the `--cov-config` option, and then one of my tests changes to a different directory and starts a subprocess, then an exception occurs during processing of the pytest-cov.pth file:

```
  File "/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 525, in <module>
    main()
  File "/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 508, in main
    known_paths = addsitepackages(known_paths)
  File "/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 288, in addsitepackages
    addsitedir(sitedir, known_paths)
  File "home/embray/miniconda/envs/test/lib/python2.6/site.py", line 185, in addsitedir
    addpackage(sitedir, name, known_paths)
  File "/home/embray/miniconda/envs/test/lib/python2.6/site.py", line 155, in addpackage
    exec line
  File "<string>", line 1, in <module>
  File "<string>", line 4, in <module>
  File "/home/embray/miniconda/envs/test/lib/python2.6/site-packages/pytest_cov/embed.py", line 58, in init
    auto_data=True)
  File "/home/embray/miniconda/envs/test/lib/python2.6/site-packages/coverage/control.py", line 138, in __init__
    "Couldn't read '%s' as a config file" % config_file
coverage.misc.CoverageException: Couldn't read 'astropy_helpers/tests/coveragerc' as a config file
```

This is because the relative path is written into the `COV_CORE_CONFIG` environment variable. So after changing directories that file cannot be found and we get a crash upon trying to process it.

On most Python versions this exception is ignored in site.py (so the coverage feature is still broken, but otherwise the tests pass).  On Python 2.6, however, the exception is not ignored and the tests fail too.

A simple workaround would be to always store the path to the config file as an absolute path.

ionelmc:
Use absolute paths for coverage config file and sources. Add a new internal variable "COV_CORE_DATAFILE" to workaround the misplacement of .coverage.foobar.123 files.

Because needs to be an absolute path COV_CORE_CONFIG we need to do some special things: coverage treats config_file='.coveragerc' as if it was config_file=True. We need to keep that behavior.

Closes #94. Closes #77.

### DIFF
--- a/src/pytest_cov/embed.py
+++ b/src/pytest_cov/embed.py
@@ -41,7 +41,8 @@ def init():
 
     cov_source = os.environ.get('COV_CORE_SOURCE')
     cov_config = os.environ.get('COV_CORE_CONFIG')
-    if cov_config:
+    cov_datafile = os.environ.get('COV_CORE_DATAFILE')
+    if cov_datafile:
         # Import what we need to activate coverage.
         import coverage
 
@@ -50,12 +51,17 @@ def init():
             cov_source = None
         else:
             cov_source = cov_source.split(os.pathsep)
+        if not cov_config:
+            cov_config = True
 
         # Activate coverage for this process.
-        cov = coverage.coverage(source=cov_source,
-                                data_suffix=True,
-                                config_file=cov_config,
-                                auto_data=True)
+        cov = coverage.coverage(
+            source=cov_source,
+            data_suffix=True,
+            config_file=cov_config,
+            auto_data=True,
+            data_file=cov_datafile
+        )
         cov.load()
         cov.start()
         cov._warn_no_data = False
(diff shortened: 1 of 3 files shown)


embray:
I can confirm that this PR fixes #94 for me.  Thanks for the quick fix!

schlamar:
Great job! :cake:

ionelmc:
The plan is to make PyPI release tomorrow.

ionelmc:
Released in `v2.2.0`.

PR #96, opened 2015-10-03
Fix handling for --cov-report=annotate.

### LINKED ISSUE #92: AssertionError when generating annotations
coverage==3.7.1
pytest==2.8.0
pytest-cov==2.1.0

I run `py.test --cov-report annotate --cov=<module> <py file>` and I get

```
  File "/Users/c17r/.virtualenvs/<project>/lib/python2.7/site-packages/pytest_cov/plugin.py", line 172, in pytest_terminal_summary
    assert total is not None, 'Test coverage should never be `None`'
_pytest.assertion.reinterpret.AssertionError: Test coverage should never be `None`
```

The annotated source is created with `<py file>,cover` so it's usable but the stack is big enough (and my window small enough) that I have to keep scrolling up to see test results

if I run `--cov-report term-missing` instead then there is no stack trace but, obviously, no annotations

ionelmc:
Add workaround for missing percentage total return from Coverage.annotate.

Also add missing test for --cov-report=annotate.

Closes #92.

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -91,7 +91,8 @@ class CovController(object):
 
         # Produce annotated source code report if wanted.
         if 'annotate' in self.cov_report:
-            total = self.cov.annotate(ignore_errors=True)
+            self.cov.annotate(ignore_errors=True)
+            total = self.cov.report(ignore_errors=True, file=StringIO())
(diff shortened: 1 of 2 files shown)


ionelmc:
Released in `v2.2.0`.

Issue #101, opened 2015-12-16
Add note to docs about pytest-cov and pytest doctest support


hangtwenty:
A coworker of mine was shocked that the built-in pytest doctest support module didn't integrate with pytest-cov. He wrote a shim to get things working. I was shocked too: I am pretty sure  things already work, if you use the arguments in the right combination.

Seems like the gist is just to make sure you have "--doctest-modules" and "--cov" arguments, in your command (or config file).

like ...

```bash
py.test foo --cov=foo --doctest-modules
```

(*EDIT* -- my previous command example WORKED but it was misleading as to what was going on. command should now be more clear.*)

I think he was forgetting `--cov` argument and confusing himself. A hint in the docs might help folks avoid this blunder going forward.

So maybe it would be good to just include a usage example like the above, in some docs somewhere. When you look for "doctest" in the pytest-cov docs you don't find anything. Maybe this little FAQ or footnote should be stuffed somewhere, so that searching the pytest-cov docs for "doctest" brings up a hint.

Happy to make a PR... just unclear where to stick this.

hangtwenty:
Could be related: https://github.com/pytest-dev/pytest-cov/issues/98

ionelmc:
Hey. So having something like "pytest-cov will work with doctests if they are in docstrings and `--cov` (or coverage configuration) matches the containing module"?

ionelmc:
I'm not sure how it even works (maybe pytest does some fancypants trickery to exec the code). We should have some tests, just in case.

hangtwenty:
Yes, something quick like you mentioned ... maybe just a bit more detailed or search-able:

> pytest-cov will work with pytest's built-in [doctest support](https://pytest.org/latest/doctest.html). If you're having trouble, ensure the argument you pass to `--cov` matches the argument you pass to `--doctest-modules`. (Or, ensure the same idea, in your configuration files.) For example, if you have a module `foo` that contains your application code, and that code has some doctests, `py.test --doctest-modules=foo --cov=foo` will include coverage data from doctests in `foo` (as well as normal coverage data).

hangtwenty:
As for fancypants trickery ...

[...]

PR #102, opened 2016-01-01
Docs: note pytest-cov and pytest doctest support

hangtwenty:
Addresses #101


ionelmc:
Sorry for this but it would be incorrect info, see comment in #101.

Issue #108, opened 2016-03-05
Using --cov-fail-under ends with traceback.


justanr:
I'm using tox to run my tests and this is happening on 2.7, 3.4, 3.5 and 3.6.0a0, [tox.ini file](https://gist.github.com/justanr/3cbc4dcb136638a2215b). To reproduce, run any suite of tests with less than the desired line coverage. [Sample traceback](https://gist.github.com/justanr/3cbc4dcb136638a2215b#file-traceback)

I only noticed this because I'm starting a new project with one test to see if I could get a combined HTML coverage report for all environments.

I'm wondering if [this line](https://github.com/pytest-dev/pytest-cov/blob/ece248242b21786c241a47f24c7a8f8f0976e735/src/pytest_cov/plugin.py#L175) should be a `sys.exit(1)` (or similar) instead

ionelmc:
Hmmm ... indeed. We could maybe create a "fake test" that fails if coverage ain't met.

The-Compiler:
It's kind of a hack too, but you could raise `pytest.UsageError` for a nicer traceback.

justanr:
Hm, `raise UsageError(...)` or `sys.exit(1)` causes tox to complain an InvocationError, but a regular run of pytest is just fine. This *could* be because of a completely local install -- not sure about that. But no traceback is output.

The-Compiler:
That sounds like the expected behaviour?

justanr:
Oh duh, that's right. I'll submit a PR for this.

ionelmc:
> It's kind of a hack too, but you could raise pytest.UsageError for a nicer output.

AFAIK someone else asked about this "hard cov-fail-under error". You think a fake test might cause problems? I wonder what @schlamar thinks about this.

schlamar:
It might be possible to get the coverage percentage in `sessionfinish`. In this case we should be able to set `session.testsfailed` to True which should set the expected error code of the process (https://github.com/pytest-dev/pytest/blob/a912d3745bcbbc49999dc9becf0d5545849d44ea/_pytest/main.py#L127)

schlamar:
I have looked further into it. We should add a `get_coverage_percentage` to the controller. There we should be able to get the results with `self.cov.report(ignore_errors=True, file=StringIO())` so we don't need the terminal writer. 

This can be called in `pytest_sessionfinish` and we can set `session.testsfailed` accordingly. 

[...]

PR #109, opened 2016-03-05
Correct hard fail with --cov-fail-under

justanr:
See #108 for discussion. I was unsure of what exactly to add to the CHANGELOG so I let it be.

I looked into creating a fake test (ala pytest-pep8), but I was unsure how to implement this.


ionelmc:
Oh my, I forgot about this. Sorry for delay and thanks!

justanr:
No issue, I forgot as well.

PR #111, opened 2016-04-04
Add Python 3.5 to setup.py

hugovk:
3.5 is missing from the badge at https://github.com/pytest-dev/pytest-cov/blob/master/README.rst but looks like you support it.


ionelmc:
Ah yes, thanks!

Issue #97, opened 2015-10-04
Allow to specify Coverage XML output file


blueyed:
I have not looked to deep, but only found this:

```
src/pytest_cov/engine.py
108: stream.write('Coverage XML written to file %s\n' % self.cov.config.xml_output)
```

Therefore it appears that the output path can only be configured through the `.coveragerc` file.

It would be nice if this could be passed in through an option to pytest-cov, e.g. `--cov-xml-output=coverage2.xml`.

ionelmc:
That would overlap a bit with the `--cov-report=xml` option. Maybe `--cov-report=xml` should be implied if you use `--cov-xml-output=coverage2.xml`. 

What if we'd be allowed to use `--cov-report=xml:coverage2.xml`?

The-Compiler:
> What if we'd be allowed to use `--cov-report=xml:coverage2.xml` (or similar)?

That sounds good, and something similar could also be done for `--cov-report=html:htmlcov2`

blueyed:
> `--cov-report=xml:coverage2.xml`

:+1:

ionelmc:
Actually this was fixed in https://github.com/pytest-dev/pytest-cov/pull/113.

PR #113, opened 2016-04-24
Specify file output

unholysampler:
Add support for specifying output location for html, xml, and annotate report. Fixes issue #97.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -13,3 +13,4 @@ Authors
 * Ionel Cristian Mărieș - http://blog.ionelmc.ro
 * Christian Ledermann - https://github.com/cleder
 * Alec Nikolas Reiter - https://github.com/justanr
+* Patrick Lannigan - https://github.com/unholysampler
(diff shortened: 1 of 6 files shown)


unholysampler:
Installed python2.6 locally and am trying to figure out why the following check is failing, even though it works in other versions and the output is the same.

```
    result.stderr.fnmatch_lines([
        '*argument --cov-report: output specifier not supported for: "term:%s"*' % DEST_DIR,
    ])
```

ionelmc:
I suspect the arg parser used there (on 2.6) has some issues ... 

You could have something like this in the tests:
```
    result.stderr.fnmatch_lines([
        '*argument --cov-report: *',
    ] if tuple(sys.python_version[:2]) == (2, 6) else [
        '*argument --cov-report: output specifier not supported for: "term:%s"*' % DEST_DIR,
    ])
```

ionelmc:
Note that we have to support 2.6, at least on some level, because pytest still supports 2.6.

ionelmc:
Thanks!

PR #115, opened 2016-05-10
Add support for 'skip_covered' option in terminal report section

### LINKED ISSUE #114: Add support for 'skip_covered' under reporting
coverage 4.0 added support for `skip_covered` option in the `[report]` section.

> `skip_covered` (boolean, default False): Don’t include files in the report that are 100% covered files. See Coverage summary for more information.[1] 

This lib doesn't seem to recognise and pass this parameter to the coverage runner.

Best. 

[1] https://coverage.readthedocs.io/en/coverage-4.0.3/config.html#report

theskumar:
This is a working implementation and works fine with `.coveragerc` and `setup.cfg`.

I've not added relevant docs and test, as i'm still trying to understand those. 

@ionelmc if this looks good, i'll try to add docs and tests. 

Ref: http://coverage.readthedocs.io/en/latest/api_coverage.html#coverage.Coverage.report

closes #114

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -16,3 +16,4 @@ Authors
 * Patrick Lannigan - https://github.com/unholysampler
 * David Szotten - https://github.com/davidszotten
 * Michael Elovskikh - https://github.com/wronglink
+* Saurabh Kumar - https://github.com/theskumar
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,8 @@ Changelog
   Contributed by David Szotten in `PR#116 <https://github.com/pytest-dev/pytest-cov/pull/116>`_.
 * Fixed bug occurred when bare ``--cov`` parameter was used with xdist.
   Contributed by Michael Elovskikh in `PR#120 <https://github.com/pytest-dev/pytest-cov/pull/120>`_.
-
+* Add support for ``skip_covered`` and added ``--cov-report=term-skip-covered`` command
+  line options. Contributed by Saurabh Kumar in `PR#115 <https://github.com/pytest-dev/pytest-cov/pull/115>`_.
 
 2.2.1 (2016-01-30)
 ------------------
--- a/README.rst
+++ b/README.rst
@@ -223,6 +223,19 @@ The terminal report with line numbers::
     --------------------------------------------------
     TOTAL                  353     20    94%
 
+The terminal report with skip covered::
+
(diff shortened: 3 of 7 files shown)


ionelmc:
Looks fine. If you need any help with the tests just ping me. (I suspect you can do something like `test_coveragerc`)

theskumar:
Thanks. 

I've updated the PR with 1) compatibility with older version of coverage 2) tests 3) docs. 

@ionelmc It would be great if you can have look & squash/merge/release.

Best.

ionelmc:
One question tho, sorry for bringing this up so late: Shouldn't we allow users to have skip_covered option when they use `--cov-report=term-missing`? In that case we'd only need to change a bit the option  parser to allow stuff like `--cov-report=term-missing:skip-covered` and `--cov-report=term:skip-covered`. What do you think?

theskumar:
so currently this is achievable with
```
--cov-report=term-missing --cov-report=term-skip-covered
```
I agree that the suggested api is cleaner, I just when the existing style... i.e. `term`, `term-missing`. 
Happy to change it, if you prefer.

ionelmc:
I think it's better for the users, not obvious that `--cov-report=term-missing` and `--cov-report=term-skip-covered` actually compose, at least not in my head.

theskumar:
Agreed. I'll take a stab at changing it.

One thing i didn't understand while working on this is: how does the `skip_covered` setting in `.coveragerc` started to get recognised by only adding an addition parameter to `self.cov.report(...)`.

ionelmc:
I suspect it's loaded when the cov object is created (and `.coveragerc` is loaded). We should have a test for this, just in case something regresses.

ionelmc:
@theskumar hey, still have time for this?

theskumar:
So i've been looking into it, and it's take more time than I aspected. I guess, mostly in the validate_report function. Happy to have someone take over this feature. I'm bit occupied as well, so just getting only short bursts of time to work.

theskumar:
@ionelmc I've updated the cli usages as suggested by you. And separated the `.coveragerc` test.

Let me know, if anything else needs to be done for this to be merged. Thanks.

ionelmc:
Looks good, thanks. Gonna wait just a bit for Travis to finish.

theskumar:
test failing :/ taking a look and will get back.

theskumar:
Updated PR with fix. Pretty sure, this will not fail!

ionelmc:
Alright, thanks for the changes.

PR #116, opened 2016-05-11
Show errors on insufficient cov

davidszotten:
don't hide failed tests when cov-fail-under fails

c.f. #108 

also tweaks failure semantics slightly, see changelog (hence minor version bump)

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -224,11 +224,11 @@ class CovPlugin(object):
         if self._should_report() and self._failed_cov_total():
             markup = {'red': True, 'bold': True}
             msg = (
-                'Required test coverage of %d%% not '
-                'reached. Total coverage: %.2f%%'
+                '\nFAIL Required test coverage of %d%% not '
+                'reached. Total coverage: %.2f%%\n'
                 % (self.options.cov_fail_under, self.cov_total)
             )
-            terminalreporter.write(msg + '\n', **markup)
+            terminalreporter.write(msg, **markup)
 
     def pytest_runtest_setup(self, item):
         if os.getpid() != self.pid:
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -1,9 +1,5 @@
 """Coverage plugin for pytest."""
 import os
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
 
 import pytest
 import argparse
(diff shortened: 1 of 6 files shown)


tests/test_pytest_cov.py
```diff
@@ -8,7 +8,7 @@
 import py
 import pytest
 import virtualenv
-from process_tests import TestProcess
```
ionelmc: Did this interfere with the default test collection settings?

src/pytest_cov/plugin.py
```diff
@@ -180,30 +184,55 @@ def pytest_testnodedown(self, node, error):
...
+            if self._failed_cov_total():
+                # make sure we get the EXIT_TESTSFAILED exit code
+                compat_session.testsfailed += 1
+
+    def pytest_terminal_summary(self, terminalreporter):
+        if self.cov_controller is None:
+            return
+
+        if self.cov_total is None:
+            # we shouldn't report, or report generation failed (error raised above)
```
ionelmc: We should still have the old assertion no? Or `pytest_terminal_summary` is called regardless of `pytest_runtestloop` raising an error? When does that happen?

tests/test_pytest_cov.py
davidszotten: this is just to suppress the warnings (mentioned in [the commit message](https://github.com/davidszotten/pytest-cov/commit/3dbb47a28999346abade322cfd9fbc0c2dcec77b), but i agree that's not very visible). would you prefer a code comment?

ionelmc:
Hey, thanks for this improvement! I'm adding couple questions on the diff.

src/pytest_cov/plugin.py
davidszotten: summary is called regardless, so we get to this comment in the two cases indicated

davidszotten:
thanks. replied inline

looks like one of the tests fails on cov 3.7 due to cov silently ignoring non-existent sources. updating to match cov behaviour by version

ionelmc:
I think this is quite fine. I wonder if messing with `session.testsfailed`/`session._testsfailed` would create any problems. I'm no pytest internals expert here, @RonnyPfannschmidt you think incrementing that attribute could be a problem?

ionelmc:
Alright, merged now. Thanks for the improvement!

davidszotten:
cool, thanks!

ionelmc:
@davidszotten it looks like these cause regression (#124). I'm looking for a fix ...

It appears that `pytest_runtestloop` don't have a way to know if it's being called in a slave or in the master. Any ideas here?

davidszotten:
hi. just saw this though it looks like you found a fix already

PR #122, opened 2016-07-06
Simplify and speedup builds in AppVeyor

nicoddemus:
Hey,

I've made some changes to the AppVeyor script to speed it up and greatly simplify it. 

- Use fewer jobs, split by Python versions: AppVeyor jobs don't run in parallel,
  so there's no benefit of having a huge matrix; in fact, it usually degrades performance
  because each build in the matrix must wait for a new chance to get a new "free" place in the queue;
- Tox from any python can be used, instead of installing tox in each python version: tox
  takes care of creating the appropriate virtual environment for each version
- Install and test with pypy (bonus)
- No need for bootstrap or compiler scripts; AppVeyor already comes with all
  Python versions pre-installed, except for pypy
- No need to test against python 32 and 64 bits; pure python projects which don't
  use C extensions usually don't need to test in both platforms

The build is still running, but I estimate it will finish in less than 1 hour, instead of the usual 4 to 4:30 hours.

I created this PR because we recently discover that the `pytest-dev` account on AppVeyor share the same "slots" in the queue for all projects under it, which is annoying because if a single commit is done to `pytest-cov` AppVeyor is stuck for over 4 hours. :sweat: 

cc @The-Compiler :sweat_smile:


nicoddemus:
Meanwhile, [here's the link](https://ci.appveyor.com/project/nicoddemus/pytest-cov/build/appveyor-speedup-4) to the build on my account.

The-Compiler:
Hmm, I'm a bit surprised - why doesn't this PR show up here in the checks or at https://ci.appveyor.com/project/pytestbot/pytest-cov/history ?

ionelmc:
Hmmm, I like this. I think this will work quite fine - `coverage` seems to have wheels for everything (except coverage 3.7 don't have wheels for py3.5 but that should be fine, no special setup needed there I think).

ionelmc:
Thanks @nicoddemus!

nicoddemus:
@ionelmc it seems the [pypy](https://ci.appveyor.com/project/pytestbot/pytest-cov/build/job/sxj4af7oh6x21bd2) build is broken... previously AppVeyor was not running pypy anyways, so I'm not sure if you want to fix the build or stop building with pypy on Windows like before. If the former, it is just a matter of removing the appropriate `TOXENV` from `ci/templates/appveyor.yml`.

nicoddemus:
> Hmm, I'm a bit surprised - why doesn't this PR show up here in the checks

Yeah, don't know about that either. :frowning:

PR #125, opened 2016-07-12
Fix definition of `cov` fixture for pytest 3

blueyed:
With pytest 3.0 (not released yet) the following warning gets emitted:

> WC1 None pytest_funcarg__cov: declaring fixtures using
> "pytest_funcarg__" prefix is deprecated and scheduled to be removed in
> pytest 4.0.  Please remove the prefix and use the @pytest.fixture decorator
> instead.

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -254,10 +254,9 @@ class CovPlugin(object):
             self.cov = None
 
 
-def pytest_funcarg__cov(request):
-    """A pytest funcarg that provides access to the underlying coverage
-    object.
-    """
+@pytest.fixture
+def cov(request):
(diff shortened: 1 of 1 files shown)


src/pytest_cov/plugin.py
```diff
@@ -254,7 +254,8 @@ def pytest_runtest_teardown(self, item):
             self.cov = None
 
 
-def pytest_funcarg__cov(request):
+@pytest.fixture
+def cov(request):
     """A pytest funcarg that provides access to the underlying coverage
```
The-Compiler: Might make sense to do funcarg -> fixture here too

blueyed: Thanks, done.

ionelmc:
Allright, Thanks!

PR #132, opened 2016-09-08
Fix and improve help for --cov-report

blueyed:
`--cov-report=` to not output any error is hard to guess.


ionelmc:
Neat argparse trick! But maybe we should make a more explicit `--cov-report=none` option? What do you think?

blueyed:
> Neat argparse trick! 

It is mentioned in the README, too.

`--cov-report=none` sounds good.
I can image most people are using it only to generate the `.coverage` file, e.g. on Travis, and then use other services on top.  But making `none` the default is probably a bit too much?!

ionelmc:
Well I was just asking for opinion. I'm going to merge this meanwhile.

PR #135, opened 2016-10-04
Implementing no_cov feature for #131

kozmaz87:
Please let me know what you think.

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -68,6 +68,9 @@ def pytest_addoption(parser):
     group.addoption('--no-cov-on-fail', action='store_true', default=False,
                     help='do not report coverage if test run fails, '
                          'default: False')
+    group.addoption('--no-cov', action='store_true', default=False,
+                    help='Disable coverage report completely (useful for debuggers) '
+                         'default: False')
     group.addoption('--cov-fail-under', action='store', metavar='MIN', type=int,
                     help='Fail if the total coverage is less than MIN.')
     group.addoption('--cov-append', action='store_true', default=False,
@@ -131,11 +134,16 @@ class CovPlugin(object):
         self.cov_total = None
         self.failed = False
         self._started = False
+        self._disabled = False
         self.options = options
 
         is_dist = (getattr(options, 'numprocesses', False) or
                    getattr(options, 'distload', False) or
                    getattr(options, 'dist', 'no') != 'no')
+        if options.no_cov:
+            self._disabled = True
+            return
+
         if is_dist and start:
             self.start(engine.DistMaster)
         elif start:
@@ -144,6 +152,7 @@ class CovPlugin(object):
         # slave is started in pytest hook
 
     def start(self, controller_cls, config=None, nodeid=None):
+
         if config is None:
             # fake config option for engine
             class Config(object):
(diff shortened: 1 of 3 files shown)


src/pytest_cov/plugin.py
```diff
@@ -226,6 +244,13 @@ def pytest_runtestloop(self, session):
                 compat_session.testsfailed += 1
 
     def pytest_terminal_summary(self, terminalreporter):
+        if self._disabled:
+            markup = {'red': True, 'bold': True}
+            msg = (
+                'WARNING: Coverage disabled by user!'
```
ionelmc: I can only assume that you make this extra cruft here to make a longer message and use the pytest's warn system (in addition to writing to terminalreporter, of course)?

If so, you can use something like `terminalreporter.config.warn(code='COV-U1', message=msg)`.

Also, I would mention "--no-cov" somehow in the warning message, just to be clear what's going on.

kozmaz87: I just copied from elsewhere in the file. It uses the same technique

ionelmc: Ah yes, you're right. Still, I think it's good if we'd use the config.warn thing here ...

kozmaz87: Also the reason for not mentioning the no-cov flag in there is precisely to avoid spaghetti coding. You would constantly have to maintain this line whenever you decide to rename this variable. Unfortunate enough already that I have to copy it in the README.rst I know that Ctrl+R exists but some ppl. prefer typing. Not to mention --no-cov vs. no_cov. By all means I will make the changes but wanted you to know the reason just for the record.

ionelmc: I'm pretty sure we can grep/sed for `--no-cov` if we ever decide to rename 😀

kozmaz87: Like I said. Your request is valid but my thought process is valid too. That's all I wanted to point out :)

kozmaz87: Also why config.warn and not pytest_logwarning

kozmaz87: I can't search for the warning line. stderr us empty and stdout does nto have it.
At this point I have no idea. I need to find another way to check for this

kozmaz87: This is for the test

kozmaz87: How can I test this? Now the log line is simply not there in the stderr/stdout?

ionelmc: I would send that message both to the warner and terminalwriter.

To test the warning you need to configure pytest to show warnings (add `-rw` to commandline).

kozmaz87: Why would you duplicate the message? Also it writes on the console for me if I run py.test without any flags but the test framework does not pick those lines up. Just 

         assert 'WCOV-U1 None Coverage disabled via --no-cov switch!' in  result.stdout.str()
    E   assert 'WCOV-U1 None Coverage disabled via --no-cov switch!' in '============================= test session starts ==============================\nplatform darwin -- Python 2.6.8 -- p...D\ntest_no_cov.py::test_foo[9] PASSED\n\n==================== 10 passed, 1 warnings in 0.02 seconds ====================='
    E    +  where '============================= test session starts ==============================\nplatform darwin -- Python 2.6.8 -- p...D\ntest_no_cov.py::test_foo[9] PASSED\n\n==================== 10 passed, 1 warnings in 0.02 seconds =====================' = <bound method LineMatcher.str of <_pytest.pytester.LineMatcher instance at 0x1050df248>>()

kozmaz87: Adding the -rw flag to the test does not help

ionelmc:
Merged and released `v2.4.0`, thanks!

PR #140, opened 2016-11-13
add option --cov-summary to show coverage summary in terminal

flniu:
In our case, we want to see a brief summary about the total coverage and required coverage in CI console, instead of printing a full term report or opening the html report. So I add an option `--cov-summary` to output coverage summary in terminal.

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -76,6 +76,9 @@ def pytest_addoption(parser):
     group.addoption('--cov-append', action='store_true', default=False,
                     help='do not delete coverage but append to current, '
                          'default: False')
+    group.addoption('--cov-summary', action='store_true', default=False,
+                    help='show coverage summary in terminal, '
+                         'default: False')
 
 
 @pytest.mark.tryfirst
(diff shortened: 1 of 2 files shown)


ionelmc:
Wouldn't an addition to `--cov-report` be better? Eg: `--cov-report=summary`.

flniu:
At first, I thought about `--cov-report=term-summary`, however:

* The summary only includes total coverage and required coverage(optional); it's not much like a report.
* I need to consider existing report types and handle the logic if no other `--cov-report` is specified.

So I choose an easier way which add an option but has the least impact on existing logic.

ionelmc:
My two concerns are:

* it doesn't really make sense to use both `--cov-summary` and `--cov-report`, as reports would show the total. The seem exclusive to me, that's why I suggested extending `--cov-report`.
* I have a preference for not adding lots of switches and knobs. Perhaps always showing some sort of summary is acceptable (eg: if there's no `--cov-report=term*`)?

flniu:
Well, I agree not to add lots of switches.

I thought about our requirement again. In current version, when `--cov-fail-under` is specified and required coverage is not reached, a message like this will be output:

>FAIL Required test coverage of 80% not reached. Total coverage: 75%

What we need is: when required coverage is reached, also output a message like:

>Required test coverage of 80% reached. Total coverage: 85%

How about this plan?

ionelmc:
If you mean changing ` --cov-fail-under ` to always show summary and adding no other option then I like it.

flniu:
Yes, that is what I mean. I'll work on it. This pr can be closed then.

PR #141, opened 2016-11-15
show total coverage when required test coverage reached

flniu:
This pr is the updated version of #140.

In our case, we specify `--cov-report=html` and `--cov-fail-under=85`, we want to see the total coverage in CI console at a glance, instead of printing a full term report or opening the html report, even when required test coverage(85%) is reached. So if total coverage is close to required, we may take actions in advance.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,6 +80,7 @@ class GeneratePTH(Command):
                     'exec(%r)' % sh.read().replace('    ', ' ')
                 )
 
+
 setup(
     name='pytest-cov',
     version='2.4.0',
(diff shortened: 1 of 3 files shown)


ionelmc:
Can you add one test (or extra assertion, surely there's a test for the passing fail-under)?

flniu:
I added 3 assertions for `--cov-fail-under` summary message.

And I also added a blank line in `setup.py` because of flake8 warning. I don't know why this didn't break the CI before.

ionelmc:
Alright, thanks!

PR #178, opened 2017-10-28
Rebased and fixed #142 again

ionelmc:
Lets see if I got it right this time. #142


ionelmc:
@ryanhiebert @ceridwen (#130) So I've got this to work but there's a small caveat: if you use non-collocated xdist you need to have proper path aliasing in your coveragerc. I think this is acceptable and can be documented for people that want such extravagant xdist setups.

blueyed:
This (combining) appears to cause coverage to drop a lot for pytest-django.
Bisected pytest-cov in https://github.com/pytest-dev/pytest-django/pull/656.

Likely/maybe related to eager setup (`COV_CORE_SOURCE` etc - https://github.com/blueyed/pytest-django/blob/a670bb79c0a3c26f81fa5602e038feefa6c1d4d7/tox.ini#L30-L32).

blueyed:
Ok, the problem with pytest-django's tests were that the eager coverage tracking creates an uncombined file, and running `coverage combine` then erased the `.coverage` that pytest-cov combined already.

Came up with https://github.com/pytest-dev/pytest-cov/pull/228 along the way.
@ryanhiebert @ceridwen (via #130): please test if it works for you still.

PR #182, opened 2017-11-21
Only test against a single recent pytest and coverage.py

hugovk:
I'd like to suggest running 146 build jobs is a bit unweildy. It uses nearly 7 hours of CI time in total, about 1.5 hours from start to finish ([eg.](https://travis-ci.org/pytest-dev/pytest-cov/builds/294101109)).

It's testing 6 Python versions against 4 pytest versions and 6 coverage.py versions (plus 2 for docs/checks), 6 * 4 * 6 + 2=146.

* Is it really necessary to test against all those old versions of pytest and coverage.py?

* Does pytest-cov need to make promises to support so old releases?

* Does pytest promise to support 3 old releases, or coverage.py 5 old releases, or just the most recent?

Further, some three dozen of the build jobs are failing, giving less confidence in the CI/tests as a whole.

https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix says:

> *All build matrixes are currently limited to a maximum of* **200 JOBS** *for both private and public repositories. If you are on an open-source plan, please remember that Travis CI provides this service free of charge to the community. So please only specify the matrix you actually need.*

I'd recommend only testing against a single, recent version of pytest and coverage.py, 6 * 1 * 1 + 2=8, only about 22 minutes total and 6 minutes start to finish ([eg.](https://travis-ci.org/hugovk/pytest-cov/builds/305209229)). Of course, I don't know the full history, so perhaps one old extra version may be necessary, but 146 seems a bit much.

Thank you!


ionelmc:
Some fair arguments here, I'll include some of these changes (#178 already drops some stuff out).

PR #199, opened 2018-04-14
--report-type='' implies no --cov-fail-under

terencehonles:
Specifying `--cov-report=` will not output any report information and will not check minimum coverage which may be specified with `--cov-fail-under=MIN` or through the config. This change is to allow using `--cov-append` in future runs without reporting failure too early.

With this change, on the final build you will need to generate at least `--cov-report=term` or run `coverage report` directly

-------------------------

I made this change because I want to declare `fail_under` in my coverage config (for me it lives in `setup.cfg` which passes some other common options by using `addopts` (like mentioned in #195)), but my current test setup requires at least 2 passes running `pytest` which initially the sequence would be:

```sh
ENV_VAR=value pytest --cov-report=
ENV_VAR=other_value pytest --cov-append
```

However, without the proposed PR the first test fails and I would have to ignore the exit code (which is not appropriate because the test suite may have actually failed)

With this change I can leave `fail_under` as part of my coverage config instead of having to hardcode it in the CI runner or add an additional file/configuration option.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -23,3 +23,4 @@ Authors
 * Zoltan Kozma - https://github.com/kozmaz87
 * Francis Niu - https://flniu.github.io
 * Jannis Leidel - https://github.com/jezdez
+* Terence Honles - https://github.com/terencehonles
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,15 @@
 Changelog
 =========
 
+2.5.2 (2018-04-14)
+------------------
+
+* Specifying ``--cov-report=`` will not output any report information and
+  will not check minimum coverage which may be specified with
+  ``--cov-fail-under=MIN`` or through the config. This change is to allow
+  using ``--cov-append`` in future runs without reporting failure too early.
+
+
 2.5.1 (2017-05-11)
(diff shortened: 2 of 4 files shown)


ionelmc:
Hmm.. shouldn't we have a negative option, like `--cov-disable-reporting` ?

I wonder if users will trip up on this empty value (eg: they forget the `=` - `--cov-report tests.py` will have undesirable result).

terencehonles:
@ionelmc I'm up for whatever. I've actually re-worked our system to not need two test passes, but I'm willing to alter this to help whoever may need this in the future.

ionelmc:
@terencehonles agree to just close this? It doesn't sound like you or anyone needs this anymore ...

graingert:
you can actually just set:

```
ENV_VAR=value pytest --cov-fail-under=0
ENV_VAR=other_value pytest --cov-append
```

terencehonles:
@graingert are you sure that works? IIRC I tried that and the value in `setup.cfg` took precedence to that.

terencehonles:
Well it does seem like the command line flag is working now, so not sure if something changed or I didn't actually try this. Either way, if that works this PR is unnecessary

PR #224, opened 2018-09-16
tests: cleanup pytest config

blueyed:


tox.ini
```diff
@@ -38,7 +38,7 @@ deps =
 pip_pre = true
 
 commands =
-    pytest {posargs:-vv}
+    {posargs:pytest -vv}
```
blueyed: @ionelmc 
Was this done on purpose?  (or when resolving conflicts by accident)

blueyed: => https://github.com/pytest-dev/pytest-cov/pull/255

ionelmc: Oooof ... yes. I often do `tox -e something -- pip list -ol` and other things like that.

blueyed: I see.. but then you could also do `.tox/$TOXENV/bin/pip list -o`.

PR #225, opened 2018-09-16
Use coverage.Coverage directly

blueyed:


blueyed:
Canceled the build manually, after enough successes.

Issue #211, opened 2018-07-04
"Module already imported so cannot be rewritten" warning with COV_CORE_SOURCE


blueyed:
I am using the following in pytest-django's tox.ini (via the suggestion in https://github.com/pytest-dev/pytest-django/pull/412#issuecomment-340077539):

```
COV_CORE_SOURCE={toxinidir}
COV_CORE_CONFIG={toxinidir}/.coveragerc
COV_CORE_DATAFILE={toxinidir}/.coverage.eager
```

However, this requires to unset `COV_CORE_SOURCE` for the inner test runs
(which uses `runpytest_subprocess`).

I've noticed this because of a warning from pytest:

> <undetermined location>
>   Module already imported so cannot be rewritten: pytest_cov

This warning however is still triggered by the outer pytest.

It makes sense, but I wonder if there is a better way to load pytest-cov eagerly by now (mentioned in the thread with #412), or if I should just configure the outer pytest to ignore this warning?

Related issue https://github.com/pytest-dev/pytest-cov/issues/148.

PR #228, opened 2018-09-21
Avoid "Module already imported" warning from pytest

blueyed:
> Module already imported so cannot be rewritten

Fixes https://github.com/pytest-dev/pytest-cov/issues/211


ionelmc:
Should we have an integration test for this (if it's easy to do)? Otherwise I don't have anything against this.

matthew-piziak:
Will this pull request be merged?

blueyed:
Merged without a test - I do not think it is worth it.

PR #230, opened 2018-09-23
Don't use deprecated config.warn

MarSoft:
`config.warn` and `node.warn` were deprecated in pytest 3.8: 
https://docs.pytest.org/en/latest/deprecations.html#config-warn-and-node-warn

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -235,7 +235,10 @@ class CovPlugin(object):
                 message = 'Failed to generate report: %s\n' % exc
                 session.config.pluginmanager.getplugin("terminalreporter").write(
                     'WARNING: %s\n' % message, red=True, bold=True)
-                warnings.warn(pytest.PytestWarning(message))
+                if pytest.__version__ >= '3.8':
+                    warnings.warn(pytest.PytestWarning(message))
+                else:
+                    session.config.warn(code='COV-2', message=message)
                 self.cov_total = 0
             assert self.cov_total is not None, 'Test coverage should never be `None`'
             if self._failed_cov_total():
@@ -246,7 +249,10 @@ class CovPlugin(object):
         if self._disabled:
             message = 'Coverage disabled via --no-cov switch!'
             terminalreporter.write('WARNING: %s\n' % message, red=True, bold=True)
-            warnings.warn(pytest.PytestWarning(message))
+            if pytest.__version__ >= '3.8':
+                warnings.warn(pytest.PytestWarning(message))
+            else:
+                terminalreporter.config.warn(code='COV-1', message=message)
             return
         if self.cov_controller is None:
             return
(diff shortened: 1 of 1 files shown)


src/pytest_cov/plugin.py
```diff
@@ -234,7 +235,7 @@ def pytest_runtestloop(self, session):
                 message = 'Failed to generate report: %s\n' % exc
                 session.config.pluginmanager.getplugin("terminalreporter").write(
                     'WARNING: %s\n' % message, red=True, bold=True)
-                session.config.warn(code='COV-2', message=message)
+                warnings.warn(code='COV-2', message=message)
```
nicoddemus: This should probably be:

```python
warnings.warn(pytest.PytestWarning(message))
```

src/pytest_cov/plugin.py
```diff
@@ -245,7 +246,7 @@ def pytest_terminal_summary(self, terminalreporter):
         if self._disabled:
             message = 'Coverage disabled via --no-cov switch!'
             terminalreporter.write('WARNING: %s\n' % message, red=True, bold=True)
-            terminalreporter.config.warn(code='COV-1', message=message)
+            warnings.warn(code='COV-1', message=message)
```
nicoddemus: Ditto.

src/pytest_cov/plugin.py
```diff
@@ -234,7 +235,10 @@ def pytest_runtestloop(self, session):
                 message = 'Failed to generate report: %s\n' % exc
                 session.config.pluginmanager.getplugin("terminalreporter").write(
                     'WARNING: %s\n' % message, red=True, bold=True)
-                session.config.warn(code='COV-2', message=message)
+                if pytest.__version__ >= '3.8':
+                    warnings.warn(pytest.PytestWarning(message))
+                else:
+                    session.config.warn(code='COV-2', message=message)
```
blueyed: Might be good to have this pattern in the deprecation docs.

src/pytest_cov/plugin.py
```diff
@@ -1,5 +1,6 @@
 """Coverage plugin for pytest."""
 import os
+import warnings
```
blueyed: Could also be made conditional (`if pytest.__version__ >= '3.8':`).

src/pytest_cov/plugin.py
nicoddemus: Good idea, done: https://github.com/pytest-dev/pytest/pull/4029

adamantike:
@MarSoft, this fix doesn't work for PyTest `3.10`, as `'3.10.1' >= '3.8'` is false. Let me know if filing a separate issue is better for tracking this.

adamantike:
I've just filed PR #353, so we can have this same fix for PyTest 3.10 too!

Issue #247, opened 2018-12-18
Test "test_cover_looponfail" fails against pytest-3.9.0+


stanislavlevin:
Pytest 3.9.0 changed API for ``Testdir.runpytest_subprocess()``  and ``Testdir.run()``.
This resulted in:
```
========================================== FAILURES ===========================================
____________________________________ test_cover_looponfail ____________________________________
/usr/src/RPM/BUILD/python-module-pytest-cov-2.6.0/tests/test_pytest_cov.py:1041: in test_cover_looponfail
    script) as process:
/usr/lib/python2.7/site-packages/_pytest/pytester.py:914: in runpytest
    return self._runpytest_method(*args, **kwargs)
E   TypeError: <lambda>() got an unexpected keyword argument 'timeout'
=================================== short test summary info ===================================
FAIL tests/test_pytest_cov.py::test_cover_looponfail
```

Issue #239, opened 2018-11-15
CI: update matrix


blueyed:
I suggest only testing pytest 3 (`<4.0`) and above, i.e `t310` and `t40`.
Also having pytest-master in the matrix would be helpful.

What do you think?

blueyed:
Started at https://github.com/pytest-dev/pytest-cov/pull/240.
The current matrix is quite ridiculous big IMHO.

ionelmc:
I agree about supporting only 2 pytest versions. We could even only test with the latest - people that need old pytest can simply use older pytest-cov (it's not like past pytest-cov releases are horrendously broken).

PR #248, opened 2018-12-18
Fix test_cover_looponfail test

stanislavlevin:
Pytest 3.9.0 introduced a timeout for ``Testdir.runpytest_subprocess()``
and ``Testdir.run()``. So API has been changed and now it looks like
``run(self, *cmdargs, **kwargs)``.

Hence to comply with the new API it needs to fix a lambda's definition.

Fixes: https://github.com/pytest-dev/pytest-cov/issues/247
Signed-off-by: Stanislav Levin <slev@altlinux.org>


blueyed:
Thanks.
Would be good to update the CI build matrix, but that is not trivial (https://github.com/pytest-dev/pytest-cov/issues/239).

PR #253, opened 2019-01-06
Use get_closest_marker

blueyed:
Fixes pytest-cov with pytest 4.1.
Requires pytest 3.6+ now.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -121,7 +121,7 @@ setup(
         'cover', 'coverage', 'pytest', 'py.test', 'distributed', 'parallel',
     ],
     install_requires=[
-        'pytest>=2.9',
+        'pytest>=3.6',
         'coverage>=4.4'
     ],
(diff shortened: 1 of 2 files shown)


blueyed:
Would be easy to not bump the pytest requirement, but I think it is good going forward there.

hugovk:
Thanks for fixing this!

Will there be a release soon? I think a lot of projects will run into build failures if they upgrade pytest.

shashank88:
Hi, Is there a release soon for this? Our builds seem to break due to this.

hugovk:
A temporary workaround is to pin pytest < 4.1

For example:

`pip install "pytest<4.1"`

shashank88:
Yeah, the backup plan was to revert back to 4.0.2 (last successful version for us) if the release isn't due soon.

ionelmc:
pytest-cov 2.6.1 is released.

shoyer:
thanks @ionelmc !

Issue #257, opened 2019-01-08
huge dependency change between 2.6.0 and 2.6.1


smarie:
In 2.6.1 pytest version >=3.6 is required, it was not the case in 2.6.0.

This made three of my [pytest plugin projects](https://github.com/smarie/OVERVIEW#tests) builds fail simultaneously yesterday :) indeed their Travis build matrix tests compliance with pytest 2.x and 3.x. 

Of course it is not a bad thing to use the latest versions available, my comment is just about the very hazardous choice of using a "patch" version number increment for such huge change.

(No action is required from this ticket, I now use hardcoded `pytest-cov` versions in my builds. You can close this after reading :) )

blueyed:
Thanks for letting us know.
This was done in https://github.com/pytest-dev/pytest-cov/pull/253, and it wouldn't have been hard to not require this after all.
So if you think it's worth it you could revert this bump and support the old API with older pytest versions I guess.

ionelmc:
Dully noted, we'll do a minor version change when we mess with constraints.

But one has to wonder why someone would bump pytest-cov but not pytest's version :)

smarie:
Thanks for the fast feedback !

blueyed:
https://github.com/pytest-dev/pytest-cov/pull/264 reverts this.

PR #264, opened 2019-02-11
Allow pytest 3.0 again

blueyed:
Ref: https://github.com/pytest-dev/pytest-cov/issues/257

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -295,7 +295,7 @@ class CovPlugin(object):
 
     @compat.hookwrapper
     def pytest_runtest_call(self, item):
-        if (item.get_closest_marker('no_cover')
+        if (compat.get_closest_marker(item, 'no_cover')
                 or 'no_cover' in getattr(item, 'fixturenames', ())):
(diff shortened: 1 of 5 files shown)


src/pytest_cov/plugin.py
```diff
@@ -291,15 +292,26 @@ def pytest_runtest_teardown(self, item):
...
+    else:
+        @compat.hookwrapper
+        def pytest_runtest_call(self, item):
+            if (item.get_marker('no_cover')
+                    or 'no_cover' in getattr(item, 'fixturenames', ())):
+                self.cov_controller.pause()
+                yield
+                self.cov_controller.resume()
+            else:
+                yield
```
blueyed: Could also be moved to compat maybe, fix `functools.partial` maybe.

ionelmc:
@blueyed do we really want this? I don't recall anyone really demanding support for older pytest :)

blueyed:
I thought it was easier.
It was via https://github.com/pytest-dev/pytest-cov/issues/257.

ionelmc:
I think we can drop this, there's no demand.

PR #265, opened 2019-02-16
Remove multiprocess finalizer and improve subprocess docs

### LINKED ISSUE #250: Pytest-cov leaves .coverage.hostname.number.number data files if running tests against multiprocessing.Pool
Running pytest with coverage for tests against multiprocessing.Pool generates several data files, but not all of them are cleaned after. It looks similar to #100.
`Pool.join()` doesn't seem to change anything.
Example filename `.coverage.myhost.local.7582.948066`

Here's the repository to reproduce: https://github.com/manycoding/pytest_cov_pool_datafiles_250

**Your operating system name and version:**
Mac OS Mojave 10.14 (18A391)

**Any details about your local setup that might be helpful in troubleshooting:**

    pipenv shell; pipenv install

Pipfile
```
pytest = "*"
pytest-cov = "*"
pytest-mock = "*"
tox-pipenv = "*"
pytest-pythonpath = "*"
```
**Detailed steps to reproduce the bug:**

    pytest --cov=src --cov-report=term-missing  tests/test_cov_pool.py

ionelmc:
Should also fix #250.


docs/subprocess-support.rst
```diff
@@ -0,0 +1,118 @@
+==================
+Subprocess support
+==================
+
+Although pytest-cov supports subprocesses and multiprocessing. However, there are few pitfalls that need to be
```
blueyed:
```suggestion
pytest-cov supports subprocesses and multiprocessing. However, there are a few pitfalls that need to be
```

docs/subprocess-support.rst
```diff
@@ -0,0 +1,118 @@
...
+explained.
+
+Normally coverage writes the data via a pretty standard atexit handler. However, if the subprocess doesn't exit on it's
```
blueyed:
```suggestion
Normally coverage writes the data via a pretty standard atexit handler. However, if the subprocess doesn't exit on its
```

docs/subprocess-support.rst
```diff
@@ -0,0 +1,118 @@
...
+own then the atexit handler might not run. Why that happens is best left to the adventurous to discover by waddling
+though the python bug tracker.
```
blueyed:
```suggestion
though the Python bug tracker.
```

docs/subprocess-support.rst
```diff
@@ -0,0 +1,118 @@
...
+            p.join()  # <= THIS IS ESSENTIAL
+
+.. _cleanup_on_sigterm:
+
+If you abuse ``multiprocessing.Process.terminate``
+==================================================
+
+It appears that many people are using the ``terminate`` method and then get unreliable coverage results.
+
+On Linux usually that means a SIGTERM gets sent to the process. Unfortunately Python don't have a default handler for
```
blueyed:
```suggestion
On Linux usually that means a SIGTERM gets sent to the process. Unfortunately Python doesn't have a default handler for
```

docs/subprocess-support.rst
```diff
@@ -0,0 +1,118 @@
...
+
+    try:
+        from pytest_cov.embed import cleanup_on_sigterm
+    except ImportError:
+        pass
+    else:
+        cleanup_on_sigterm()
+
+
+On Windows there's no nice way to do cleanup (no signal handlers) so you're left to your own devices.
```
blueyed:
```suggestion
On Windows there's no nice way to do cleanup (no signal handlers).
```

docs/subprocess-support.rst
```diff
@@ -0,0 +1,118 @@
...
+
+If anything else
+================
+
+If you have custom signal handling, eg: you do reload on SIGHUP you should have something like this:
```
blueyed:
```suggestion
If you have custom signal handling, e.g. you do reload on SIGHUP you should have something like this:
```

ionelmc:
So it turns out there is an actual regression. I may need to bring back the finalizers :|

tests/test_pytest_cov.py
```diff
@@ -889,6 +889,38 @@ def test_funcarg_not_active(testdir):
...
+
+    script = testdir.makepyfile('''
+import multiprocessing
+
+def target_fn(a):
+    return a + 1
+
+def test_run_target():
+    for i in range(100):
+        with multiprocessing.Pool(10) as p:
```
manycoding: Why this test doesn't use the recommended way https://github.com/pytest-dev/pytest-cov/pull/265/files#diff-c1d4de3cfc01566955163b584f0efcc0R29 ?

ionelmc: So I added support for the old way while fixing the regressions. Need to update the docs again. The new way is still the best way for the pytest-cov<=2.6.1 releases.

ionelmc:
So I'm in a bit of a dilema here: currently the PR installs a cleaning SIGTERM handler via multiprocessing.util.register_after_fork. This allows using the Pool context manager API. However, this is a bit inconsistent with everything else (we don't register a SIGTERM handler elsewhere automatically).

The signal handling is now re-entrant and calls the previous handling (or the default handling), so in theory we could always install it.

Pls vote via reactions (@manycoding @blueyed @pytest-dev/pytest-cov-admin or anyone): 
* 👍 - for always installing it (and maybe do a pytest-cov 3.0 to signal that something big is going on)
* 👎 - for not installing it at all (remove it from multiprocessing.util.register_after_fork, and force users to use `pool.close()` and such)
* 😕 - for keeping it inconsistent (like it's now in the PR)

ionelmc:
There's also the 4th option of making user opt-in into the heavy handed signal handling by using `--cov-sigterm` or something like that.

blueyed:
I think enabling it via a config option would be good, and allow to easily add it to your config (or use `-o cov_install_sighandlers`).
But given that there is no / not much overhead, I think using it by default would be helpful - and then an option would be used to turn it off.
The option could be a list of signals to handle, defaulting to all then, and setting it to an empty list would disable it then.

appveyor.yml
```diff
@@ -8,7 +8,7 @@ environment:
     - TOXENV: 'py27-t310-c45,py27-t40-c45,py27-t41-c45'
     - TOXENV: 'py34-t310-c45,py34-t40-c45,py34-t41-c45'
     - TOXENV: 'py35-t310-c45,py35-t40-c45,py35-t41-c45'
-    - TOXENV: 'pypy-t310-c45,pypy-t40-c45,pypy-t41-c45,pypy3-t310-c45,pypy3-t40-c45,pypy3-t41-c45'
+    - TOXENV: 'pypy-t310-c45,pypy-t40-c45,pypy-t41-c45'
```
blueyed: I think it could be kept, but the problem appears to be that only pypy gets installed (unconditionally)?!

blueyed: See https://github.com/pytest-dev/pytest/commit/ec69514eb2be687f751e3a7a18d96e3e29d44c16 for installing it only if needed.

ionelmc: We don't install pypy there at all - it's preinstalled. Either way, I don't want to bother with pypy3 yet on Windows (it's beta anyway, and xdist is broken on pypy windows, so why bother with extensive testing on such a platform).

ionelmc:
So my concern is that the signal handling will introduce side-effects. There's already a strange problem with the pypy tests on travis, they often get stuck on test_multiprocessing_pool_terminate. Multiprocessing support is already broken on Windows due to various platform limitations or python bugs.

So now that I think of it more it should definitely be opt-in - so users notice the limitations when they look how to get the extra cleanup functionality.

ionelmc:
So I think we can do without yet another config/cli option for now. Check the new doc changes, they should be the final ones, short of grammar/typo issues.

ionelmc:
Well the merge commit message is a bit wrong but so is my hastiness 😵

PR #266, opened 2019-02-16
Add "testing" extras_require, use it in tox

blueyed:


ionelmc:
@blueyed does anyone actually need this change? Cause we might need to revert this as latest xdist wants pytest>=4.4

blueyed:
@ionelmc 
It makes it easier to setup the testing env - maybe keep it without a version, and pin it in tox then?
(IIRC you/somebody wanted it pinned)

ionelmc:
Hmmm it could work. Right now xdist is removed from setup.py - I'll try to add it back without pin.

Issue #234, opened 2018-11-05
Coverage report generation incompatible with pytest `tmpdir` feature


bluenote10:
Given a project structure like:

```
myproject
├── myproject
│   └── __init__.py
├── tests
│   └── myproject
│       └── test.py
└── setup.cfg
```

I'm running the command in the toplevel `myproject` directory:
```
py.test --cov=myproject --cov-report html
```

On stdout I can see the log output:

```
Coverage HTML written to dir htmlcov
```

However no output has been written, i.e., the `myproject` directory does not contain `htmlcov` (and neither of its subdirectories). Even a `find / -name "*htmlcov*"` reveals nothing.

Notes:
- The output is written if I use `--cov-report html:/absolute/path/to/myproject/htmlcov`.
- I get the same issue for `--cov-report xml`.
- My tests `chdir` a lot -- can this mess up pytest-cov?

Versions:
- pytest: 3.8.1
- pytest-cov: 2.6.0

blueyed:
> My tests chdir a lot -- can this mess up pytest-cov?

Sounds like it.  I assume it gets written to some temporary directory maybe (outside of with you run `find` then).
Try exporting `COV_DEBUG` (see coverage.py doc for options) to see what is going on.

blueyed:
Is `.coverage` created as expected?
I.e. does `coverage report` work after running the tests?
Then `coverage html htmlcov` might work also from there.

bluenote10:
Yes the `.coverage` file gets created in the current directory and I also get sensible output from the terminal reporters. When running `coverage html htmlcov` afterwards in the project directory I get the error:

```
No source for code: '/home/me/workspace/myproject/htmlcov': [Errno 21] Is a directory: u'/home/me/workspace/myproject/htmlcov'
```

This at least creates the `htmlcov` folder in the project directory, but it is empty.

I haven't gotten much out of exporting `COV_DEBUG` so far.

blueyed:
btw: it is probably better to look at pytest-cov's code, where it handles the html report - i.e. put a `pdb.set_trace()` there / print `os.getcwd()`.

bluenote10:
> The command should be coverage html -d htmlcov, of course - sorry.

Ah yes, this generates the output just fine.

> What options are you using?

What I did was just `export COV_DEBUG=true` because I couldn't find any documentation on `COV_DEBUG`. I also don't have a `coverage.py`.

blueyed:
[...]

Issue #218, opened 2018-08-16
doc: example: coverage combine not needed?


blueyed:
I think the following line is not needed, is it?

https://github.com/pytest-dev/pytest-cov/blame/115155d34fc8661268c864238bc07b37938bbd0c/example/tox.ini#L29

blueyed:
It would either error out (#222), or would overwrite the `.coverage` that pytest-cov created already.

Issue #212, opened 2018-07-05
warning "Module x was never imported" when running with xdist


anthrotype:
If I install pytest-xdist plugin, and run `pytest --cov` with some `-n` number of parallel processes (even with `-n 1`), I get this warnings at the end:

```
Coverage.py warning: Module psautohint was never imported. (module-not-imported)
Coverage.py warning: No data was collected. (no-data-collected)
```

However, the data is collected, the `.coverage.*` data file is created, and a report is printed correctly despite the warnings.

If I run without passing the -n xdist option, so not in parallel, I don't get any such warnings.
Similarly I don't get those warning if I run `coverage run -m pytest` instead of `pytest --cov`.

I have a .coveragerc file in the root of the repository which contains this:

```
source = psautohint

[paths]
source =
    src/psautohint
    .tox/*/lib/python*/site-packages/psautohint
    .tox/*/lib/site-packages/psautohint
    .tox/pypy*/site-packages/psautohint
```

ionelmc:
How can I reproduce this? Enough to checkout https://github.com/adobe-type-tools/psautohint and run tox-something?

anthrotype:
thanks for looking into this. Yes, you can just clone and run `tox`, you should get a warning like this
https://travis-ci.org/adobe-type-tools/psautohint/jobs/414896624#L498

ionelmc:
Check the new examples from #267 - they should clarify how to correctly use pytest-cov in your project.

anthrotype:
Thank you!

manoadamro:
having the same problem, was working fine, now, having changed exactly nothing (i hit the up arrow and then enter) 
it's importing everything from site packages (on a different hard drive) but for all the source files it says "module was never imported", yet hundreds of tests that use that module ran and passed...

Is there any known reasons as to why I can only use coverage on windows a couple of times before it completely breaks? I have never had this issue on mac

PR #267, opened 2019-02-17
Recreate the examples

ionelmc:
Solves common user problems regarding cov data aggregation and layout.

Would need to link a bunch of issues here about aliasing/layout/config issues.


ionelmc:
Maybe I should include changes for https://github.com/pytest-dev/pytest-cov/issues/169 https://github.com/pytest-dev/pytest-cov/issues/170 https://github.com/pytest-dev/pytest-cov/issues/176

Issues that would be closed by this: https://github.com/pytest-dev/pytest-cov/issues/197 https://github.com/pytest-dev/pytest-cov/issues/218 https://github.com/pytest-dev/pytest-cov/issues/212 (it has aliasing issue) https://github.com/pytest-dev/pytest-cov/issues/234 (need to use absolute paths, like in adhoc example)

ionelmc:
So this should be complete, lemme know if I missed something.

ionelmc:
@blueyed any comments?

.travis.yml
```diff
@@ -5,74 +5,112 @@ env:
...
       env:
         - TOXENV=py27-t310-c45
-    - python: '2.7'
+    - stage: tests
+      python: '2.7'
       env:
         - TOXENV=py27-t40-c45
-    - python: '2.7'
+    - stage: tests
+      python: '2.7'
```
blueyed: You do not have to repeat the stage names - I'd split them into sections separated by a newline.
Requires extra logic in the template generator though, so likely not worth it.

The generated files (.travis.yml etc) should have a comment about them being autogenerated though btw, since I've previously missed this myself.

docs/config.rst
```diff
@@ -8,9 +8,9 @@ further control of coverage use a coverage config file.
 For example if tests are contained within the directory tree being measured the tests may be
 excluded if desired by using a .coveragerc file with the omit option set::
 
-    pytest --cov-config .coveragerc
-            --cov=myproj
-            myproj/tests/
+    pytest --cov-config=.coveragerc
+           --cov=myproj
+           myproj/tests/
```
blueyed: Have this in a single line?

.travis.yml
blueyed: https://github.com/pytest-dev/pytest-cov/pull/278

PR #268, opened 2019-03-08
[WIP] tests: use inprocess method with pytester by default

blueyed:


setup.py
```diff
@@ -129,7 +129,9 @@ def run(self):
         'testing': [
             'fields',
             'hunter',
+            # TODO: unpin?!
             'process-tests==2.0.0',
+            # TODO: unpin?!
```
ionelmc: So the idea was that things we test against (iow: we expect regressions or integration issues on those) need to be pinned so we have a predictable test suite. That way we're able to say: this pytest-cov release is compatible with pytest-xdist==X.Y.Z

If we don't have pins then things can get pretty random.

Also: https://caremad.io/posts/2013/07/setup-vs-requirement/ ('s/requirement.txt/tox.ini/`)

Pins shouldn't exist in setup.py as it can break user install scenarios.

blueyed: Those are only for extras_require=testing.

blueyed:
Now getting failures related to https://github.com/pytest-dev/pytest-cov/pull/271 (helper not importable).. will rebase/investigate later.

setup.py
blueyed: Addressed in 735c720.

ionelmc:
So I expect this will take some trial and error to get it right - some tests can't work with inprocess mode cause they need the suprocess support (wich needs the python startup hook we have in the pth file). Also, the xdist stuff might be broken (I suspect xdist messes up the process and can't be used more than one time, if it even works the first time).

blueyed:
> So I expect this will take some trial and error to get it right

Yes.

PR #269, opened 2019-03-08
ci: Travis: dist=xenial

blueyed:


blueyed:
Only found https://github.com/pytest-dev/pytest-cov/blob/master/ci/templates/.travis.yml now - I assume it should go through it instead?

ionelmc:
Yeah ... saves lots of typing.

blueyed:
Looks good, but pypy might still fail/hang: https://travis-ci.org/pytest-dev/pytest-cov/jobs/503572051

ionelmc:
I rebased it for you. That pypy test should be skipped now (xdist is very lousy on pypy).

ionelmc:
@blueyed one question tho: why `cache: false` instead of `cache: pip`?

blueyed:
@ionelmc 
Travis caches get downloaded from S3(?), which is similar to using pip via CDN and (hopefully) http caching.
This gets rid of overhead of managing the cache.

Issue #270, opened 2019-03-08
Local test failure: ModuleNotFoundError: No module named 'helper'


blueyed:
I am seeing the following test failure locally, also when using tox, even from
a fresh git clone:

```
platform linux -- Python 3.7.2, pytest-4.3.0, py-1.8.0, pluggy-0.9.0
rootdir: …/Vcs/pytest-cov, inifile: setup.cfg
plugins: forked-1.0.2, cov-2.6.1
collected 113 items

tests/test_pytest_cov.py F

=========================================================================================== FAILURES ===========================================================================================
____________________________________________________________________________________ test_central[branch2x] ____________________________________________________________________________________
…/Vcs/pytest-cov/tests/test_pytest_cov.py:187: in test_central
    '*10 passed*'
E   Failed: nomatch: '*- coverage: platform *, python * -*'
E       and: '============================= test session starts =============================='
E       and: 'platform linux -- Python 3.7.2, pytest-4.3.0, py-1.8.0, pluggy-0.9.0 -- …/Vcs/pytest-cov/.venv/bin/python'
E       and: 'cachedir: .pytest_cache'
E       and: 'rootdir: /tmp/pytest-of-user/pytest-647/test_central0, inifile:'
E       and: 'plugins: forked-1.0.2, cov-2.6.1'
E       and: 'collecting ... collected 0 items / 1 errors'
E       and: ''
E       and: '==================================== ERRORS ===================================='
E       and: '_______________________ ERROR collecting test_central.py _______________________'
E       and: "ImportError while importing test module '/tmp/pytest-of-user/pytest-647/test_central0/test_central.py'."
E       and: 'Hint: make sure your test modules/packages have valid Python names.'
E       and: 'Traceback:'
E       and: 'test_central.py:1: in <module>'
E       and: '    import sys, helper'
E       and: "E   ModuleNotFoundError: No module named 'helper'"
E       and: ''
E   fnmatch: '*- coverage: platform *, python * -*'
E      with: '----------- coverage: platform linux, python 3.7.2-final-0 -----------'
E   nomatch: 'test_central* 9 * 85% *'
E       and: 'Name              Stmts   Miss Branch BrPart  Cover   Missing'
E       and: '-------------------------------------------------------------'
E       and: 'test_central.py       9      8      4      0     8%   3-11'
E       and: ''
[...]

PR #271, opened 2019-03-08
tests: set PYTHONPATH via session scoped fixture

blueyed:
Fixes https://github.com/pytest-dev/pytest-cov/issues/270.


ionelmc:
This seems like a fine change for ppl running the suite without tox.

PR #278, opened 2019-03-25
ci: revisit templates

blueyed:
- add a comment about being auto-generated
- more compact jobs with Travis

PR #279, opened 2019-03-25
ci: pytest 4.3

blueyed:
Includes https://github.com/pytest-dev/pytest-cov/pull/278.


ionelmc:
Is it just me or the diff comes from a cache? Cause the first commit should be in master now.

ionelmc:
Well dunno, must be a github bug. Thanks for fixing the ci.

PR #281, opened 2019-03-30
[WIP/RFC] use True as default for --cov-config

blueyed:
TODO:

- [ ] refactor to be used with other controllers

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -46,9 +46,8 @@ class CovController(object):
             os.environ['COV_CORE_SOURCE'] = os.pathsep
         else:
             os.environ['COV_CORE_SOURCE'] = os.pathsep.join(self.cov_source)
-        config_file = os.path.abspath(self.cov_config)
-        if os.path.exists(config_file):
-            os.environ['COV_CORE_CONFIG'] = config_file
+        if self.abs_cov_config_file:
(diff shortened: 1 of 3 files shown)


blueyed:
@ionelmc 
Any input on this?
/cc @nedbat (https://github.com/nedbat/coveragepy/issues/788)

src/pytest_cov/engine.py
```diff
@@ -151,10 +150,20 @@ def start(self):
         self.cov = coverage.Coverage(source=self.cov_source,
                                      branch=self.cov_branch,
                                      config_file=self.cov_config)
+
+        # Get absolute location of used config file.
+        parsed_config_file = self.cov.config.config_file
+        if parsed_config_file:
+            parsed_config_file = os.path.abspath(parsed_config_file)
+            assert os.path.exists(parsed_config_file)
```
ionelmc: I think this would cause failures when coverage config is completely missing. I suspect this is hosing up all the tests.

blueyed: Tests are mostly broken because it requires coveragepy 5.

nedbat:
@blueyed I took a stab at the coverage.py change: https://github.com/nedbat/coveragepy/tree/nedbat/configfile-simplification

Running the pytest-cov tests with that, there are many failures because the config file handling has to change here.  I'm not sure it's worth it now, but maybe you want to take a stab at it?

PR #311, opened 2019-07-26
Support float values for --cov-fail-under.

### LINKED ISSUE #260: --cov-fail-under support float values

mgaitan:
It only show the decimal part if needed. Fix #260.

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -73,7 +73,7 @@ def pytest_addoption(parser):
     group.addoption('--no-cov', action='store_true', default=False,
                     help='Disable coverage report completely (useful for debuggers). '
                          'Default: False')
-    group.addoption('--cov-fail-under', action='store', metavar='MIN', type=int,
+    group.addoption('--cov-fail-under', action='store', metavar='MIN', type=float,
                     help='Fail if the total coverage is less than MIN.')
     group.addoption('--cov-append', action='store_true', default=False,
                     help='Do not delete coverage but append to current. '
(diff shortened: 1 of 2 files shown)


src/pytest_cov/plugin.py
```diff
@@ -263,20 +263,24 @@ def pytest_terminal_summary(self, terminalreporter):
 
         terminalreporter.write('\n' + self.cov_report.getvalue() + '\n')
 
-        if self.options.cov_fail_under is not None and self.options.cov_fail_under > 0:
-            if self.cov_total < self.options.cov_fail_under:
+        fail_under = self.options.cov_fail_under
+        if fail_under is not None and fail_under > 0:
```
mgaitan: btw, this could be just `if fail_under`, right?

src/pytest_cov/plugin.py
```diff
@@ -263,20 +263,24 @@ def pytest_terminal_summary(self, terminalreporter):
...
+            str_fail_under = str(
+                round(fail_under, 2) if fail_under % 1 else int(fail_under)
```
ionelmc: I wonder if using a Decimal would simplify this.

src/pytest_cov/plugin.py
ionelmc: Hmmm ... I suspect it was written like that to allow users to disable fail under by repeating the option with 0 as argument.

src/pytest_cov/plugin.py
graingert: this means that fail_under `--cov-fail-under==1.0` fails also?

src/pytest_cov/plugin.py
```diff
@@ -73,7 +73,7 @@ def pytest_addoption(parser):
     group.addoption('--no-cov', action='store_true', default=False,
                     help='Disable coverage report completely (useful for debuggers). '
                          'Default: False')
-    group.addoption('--cov-fail-under', action='store', metavar='MIN', type=int,
+    group.addoption('--cov-fail-under', action='store', metavar='MIN', type=float,
```
graingert:
```python
import ast
import numbers

def parse_fail_under(num_str):
    try:
        num = ast.literal_eval(num_str)
    except SyntaxError:
        raise ValueError
    if isinstance(num, numbers.Real):
        return num
    raise ValueError
```

how about this. So that 1.0 and 1 do not behave the same

mgaitan: that looks simple and elegant. +1.

src/pytest_cov/plugin.py
ionelmc: Ah nevermind, now I get what it does. It would display 0.0 and 0 as "0". But that seems pointless. Why not just have ` ... %s ... " % (round(self.options.cov_fail_under, 2) self.cov_total))`? Clear and simple.

tests/test_pytest_cov.py
```diff
@@ -343,6 +343,20 @@ def test_cov_min_50(testdir):
     ])
 
 
+def test_cov_min_float_value(testdir):
+    script = testdir.makepyfile(SCRIPT)
```
graingert: Would be good to get a test that actually passes/fails based on a fraction of a percentage coverage

src/pytest_cov/plugin.py
```diff
@@ -50,6 +50,12 @@ def __call__(self, parser, namespace, values, option_string=None):
 def pytest_addoption(parser):
     """Add options to control coverage."""
 
+    def fail_under(num_str):
```
ionelmc: Can we have this at module level, like the other argument type, perhaps as `parse_fail_under`? It doesn't need to be inside this closure.

PR #315, opened 2019-08-04
tox.ini: make "tox -e pypy" etc work

blueyed:
This uses internal env vars for deps, allowing to install them always
(unpinned).

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,21 +10,27 @@ envlist =
 extras = testing
 setenv =
     PYTHONUNBUFFERED=yes
+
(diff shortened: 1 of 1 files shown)


ionelmc:
Very clever trick, you should start a blog with these :)

blueyed:
> you should start a blog with these :)

I'd rather have a way to not having to use those.. ;)
(i.e. tox should allow for a fallback default to be used for non-factors)

blueyed:
Created https://github.com/tox-dev/tox/issues/1390.

PR #316, opened 2019-08-04
ci: AppVeyor: only install PyPy if used

blueyed:


blueyed:
Tried it in another PR, but failed to make it work.
Using Powershell might be easier, but needs adjustments then..

appveyor.yml
```diff
@@ -15,13 +15,12 @@ init:
...
 install:
-  # install pypy using choco (redirect to a file and write to console in case
-  # choco install returns non-zero, because choco install python.pypy is too
-  # noisy)
-  - choco install python.pypy > pypy-inst.log 2>&1 || (type pypy-inst.log & exit /b 1)
-  - set PATH=C:\tools\pypy\pypy;%PATH% # so tox can find pypy
-  - echo PyPy installed
-  - pypy --version
+  - ps:
+      if ($env:TOXENV -notmatch 'pypy') {
```
ionelmc: I suspect this failed cause newlines are gone? https://yaml-multiline.info/ something have yaml something :)

ionelmc:
@blueyed should we leave the cache on? may have caused the itertools issue anyway. Not much slower (in fact it's faster on the check env). I'd try to rerun the build to make sure about the timing difference but I can't.

blueyed:
> should we leave the cache on?

In general I'm not in favor of using pip caches on CI, but OK to keep it with me.

blueyed:
> I'd try to rerun the build to make sure about the timing difference but I can't.

You could push an empy "rebuild" commit.
But I think it's not worth it.

ionelmc:
Very well.

PR #314, opened 2019-08-04
Fix AppVeyor

blueyed:
Current failure: https://ci.appveyor.com/project/pytestbot/pytest-cov/builds/26457211/job/2nweq08e72u5m6c2

PR #317, opened 2019-08-04
ci: AppVeyor: upgrade virtualenv to fix builds

blueyed:
Taken out of https://github.com/pytest-dev/pytest-cov/pull/314.


blueyed:
While the first commit fixes the build, it uses py27 instead of pypy in the latest jobs..
Makes no sense to create a second set of tries here like in https://github.com/pytest-dev/pytest-cov/pull/314, closing.

PR #318, opened 2019-08-09
Add project structure to Usage

sylhare:
Hi,

I have added the project structure to the usage part, because I felt like it was not clear at first how to use it only based on just the command.

With the tree view, it appears more clearly (at least to me) what you need to put for it to work.
If you think that this change is 👍  maybe there are other places in the doc I need to add it? Or stuff I need to do for this pull request to be accepted.
If you thing the change is 👎  then sorry 😅 

Thanks

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -101,20 +101,37 @@ Under certain scenarios a stray ``.pth`` file may be left around in site-package
 Usage
 =====
 
+With a repository structure such as:
+
 ::
 
-    pytest --cov=myproj tests/
+   ─── repository
+       ├── myproj
+       │   ├── __init__.py
+       │   ├── feature4286.py
+       │   └── myproj.py
+       └── tests
+           ├── __init__.py
+           └── test_myproj.py
 
-Would produce a report like::
+Test the coverage of your project with:
 
-    -------------------- coverage: ... ---------------------
-    Name                 Stmts   Miss  Cover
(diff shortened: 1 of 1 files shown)


ionelmc:
A better way is to point out to https://github.com/pytest-dev/pytest-cov/tree/master/examples and not change the intro section to a project layout overview.

sylhare:
Well, the examples are not very clear, should I add an example that correspond to the intro with the command to get the same thing as the intro?
Then we can add a link toward that in the intro as you suggest.

ionelmc:
Perhaps https://github.com/pytest-dev/pytest-cov/blob/master/examples/README.rst could be edited to have an overview of commands to run, usage instructions and such.

If we add layout stuff in the main readme then we open a can of worms cause it's a too big subject to have.

Then again, I don't really understand what prompted this PR. Perhaps explain what problems you hit when you started using pytest/cov?

sylhare:
I opened this PR because it didn't look obvious to me how to use it when looking at the usage part.
Sure, I made it work, but I wanted to make it more user friendly so that even a beginner could understand it at first glance.

But as you say, we don't want to update the readme too much, so we can just add an example in `./examples/` that correspond to the usage, add some information there.

sylhare:
I guess it is not very needed, I will close it then.

PR #291, opened 2019-05-13
[WIP/waiting] ci: test with Coverage.py 5 (alpha)

blueyed:
Should we have this already, allowed to fail?
Or fix it from here?

PR #319, opened 2019-08-12
Support for coverage 5.0

nedbat:
Add support for coverage 5.0, while keeping 4.5.x working.

Looking for feedback about the approach, and help with still-failing tests.

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -114,6 +114,14 @@ class CovController(object):
             for node_desc in sorted(self.node_descs):
                 self.sep(stream, ' ', '%s' % node_desc)
 
+        # Report on any failed workers.
+        if self.failed_workers:
+            self.sep(stream, '-', 'coverage: failed workers')
+            stream.write('The following workers failed to return coverage data, '
+                         'ensure that pytest-cov is installed on these workers.\n')
+            for node in self.failed_workers:
+                stream.write('%s\n' % node.gateway.id)
+
         # Produce terminal report if wanted.
         if any(x in self.cov_report for x in ['term', 'term-missing']):
             options = {
@@ -156,14 +164,6 @@ class CovController(object):
             with _backup(self.cov, "config"):
                 self.cov.xml_report(ignore_errors=True, outfile=output)
             stream.write('Coverage XML written to file %s\n' % (self.cov.config.xml_output if output is None else output))
-
-        # Report on any failed workers.
-        if self.failed_workers:
-            self.sep(stream, '-', 'coverage: failed workers')
-            stream.write('The following workers failed to return coverage data, '
-                         'ensure that pytest-cov is installed on these workers.\n')
-            for node in self.failed_workers:
-                stream.write('%s\n' % node.gateway.id)
 
         return total
 
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -317,10 +317,6 @@ class DistWorker(CovController):
                                      branch=self.cov_branch,
                                      data_suffix=True,
                                      config_file=self.cov_config)
-        if self.cov_append:
-            self.cov.load()
-        else:
-            self.cov.erase()
         self.cov.start()
         self.set_env()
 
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -24,6 +24,7 @@ setenv =
 
     c44: _DEP_COVERAGE=coverage==4.4.2
     c45: _DEP_COVERAGE=coverage==4.5.3
+    c50: _DEP_COVERAGE=coverage==5.0a6
```
blueyed: Also add it to envlist and Travis then.
(https://github.com/pytest-dev/pytest-cov/pull/291)

nedbat: I'm not sure what the point of envlist is here.  It's the default if no "-e" is specified, but I've been told, "oh, no one runs all the environments."  Maybe there's a use of envlist I don't understand?

nedbat: Hmm, ci/bootstrap.py.  How broad a test matrix do we want at the end of this?

blueyed: `envlist` can also be seen as a list of valid/supported envs, displayed with `tox -l`.

> Hmm, ci/bootstrap.py. How broad a test matrix do we want at the end of this?

￼I personally find it too broad already.
Makes sense to only enable it explicitly while working on this PR.

ionelmc: `./ci/bootstrap.py` needs to be run after editing the envlist section.  Travis can support 200 items (maybe) in the matrix, Appveyor less (but we have some grouping there).

I think it's fine if we run both 4.5 and 5.0 for a while (pytest devs will forgive us for taking travis CI time 😇)

nedbat:
@ionelmc I've added the factor to the CI files (with bootstrap).  Can you help me understand why the tests are failing?

blueyed:
The test failure with `tests/test_pytest_cov.py::test_dist_missing_data` starts to happen with https://github.com/nedbat/coveragepy/commit/dade9db58673d4f712ba5c50ed66d1c67a7e1d5d.
The test appears to test using a virtualenv as a worker, where pytest-cov is not installed, and expects a message about it ('*- coverage: failed slaves -*'), but coverage.py does not write data anymore (https://github.com/nedbat/coveragepy/blob/dade9db58673d4f712ba5c50ed66d1c67a7e1d5d/coverage/control.py#L651), which the results in a different error.

nedbat:
This now passes all tests when run against coverage.py master.  I can make a new alpha of coverage.py to match this code if everyone is happy with it.

src/pytest_cov/engine.py
```diff
@@ -124,12 +124,12 @@ def summary(self, stream):
         # Produce html report if wanted.
         if 'html' in self.cov_report:
             total = self.cov.html_report(ignore_errors=True, directory=self.cov_report['html'])
-            stream.write('Coverage HTML written to dir %s\n' % self.cov.config.html_dir)
+            stream.write('Coverage HTML written to dir %s\n' % self.cov_report['html'])
```
graingert: this isn't quite right, because cov_report['html'] can be None

graingert: https://github.com/pytest-dev/pytest-cov/pull/330/files#diff-3db14ea221c08dabe6d00f457c4cabbeR149-R156

nedbat: You make a good point (and the test suite needs to have a test added for it).  But I'm not sure the best solution. This is part of the problem with having a plugin provide the UI for coverage: it wants to report on what happened, but doesn't have access to the internal logic. 

Before I fixed the mutation of configuration, the plugin could examine the configuration after the fact to see where the report had been written.  Now the plugin has to duplicate some logic in order to accurately show where the HTML report was written.

Should Coverage.html_report have a verbose=True parameter so that it can write the message? Should it somehow provide the final configuration to the caller?

nedbat: I've pushed a new commit to deal with this.

graingert: @nedbat I think the fix would be a function that takes a config object only

nedbat: @graingert sorry, i don't understand what that means.  A function where? That does what?

nedbat:
I've rebased this, it's ready for merging if people approve.

src/pytest_cov/engine.py
nedbat: I saw, I think I rebased on top of it correctly?

PR #321, opened 2019-08-15
update terminology to match pytest-xdist

graingert:


graingert:
new terminology was introduced in pytest-xdist==1.22.1 so run tests on pytest-xdist==1.22.0

.travis.yml
```diff
@@ -17,61 +17,89 @@ jobs:
     - env: TOXENV=docs
 
     - stage: tests
-      env: TOXENV=py27-t310-c45
+      env: TOXENV=py27-t310-c45-xdist22
```
graingert: hmm probably a tad too many envs

appveyor.yml
```diff
@@ -23,9 +23,6 @@ install:
   - echo PyPy installed
   - pypy --version
 
-  # Upgrade virtualenv for e.g. more-itertools to be handled properly.
```
graingert: oops looks like this wasn't added to the template

graingert:
I think updating the code to the `pytest-xdist>1.22.0` without breaking compatibility with `pytest-xdist<=1.22.0` will be too difficult.

.travis.yml
graingert: no actually it seems like travis copes with them fine

blueyed: ..but it's really not necessary.
And it blocks all other jobs from the pytest-dev org.

graingert: Ok I've changed the config to only add one additional test run

ci/templates/appveyor.yml
```diff
@@ -22,6 +22,9 @@ install:
   - echo PyPy installed
   - pypy --version
 
+  # Upgrade virtualenv for e.g. more-itertools to be handled properly.
+  # Pin it to work around https://github.com/tox-dev/tox/issues/1389.
+  - C:\Python35\python -m pip install -U virtualenv==16.5.0
   - C:\Python35\python -m pip install tox
```
blueyed: Yay for my code \o/ ;)
Boo for requiring this - but it is good to pin anyway.
(fix has been merged already, but not released yet: https://github.com/pypa/virtualenv/pull/1400)

docs/xdist.rst
```diff
@@ -5,9 +5,9 @@ Distributed testing (xdist)
 "load" mode
 ===========
 
-Distributed testing with dist mode set to load will report on the combined coverage of all slaves.
-The slaves may be spread out over any number of hosts and each slave may be located anywhere on the
-file system.  Each slave will have its subprocesses measured.
+Distributed testing with dist mode set to load will report on the combined coverage of all workers.
```
blueyed:
```suggestion
Distributed testing with dist mode set to "load" will report on the combined coverage of all workers.
```

tox.ini
```diff
@@ -1,9 +1,10 @@
-; a generative tox configuration, see: https://testrun.org/tox/latest/config.html#generative-envlist
+; a generative tox configuration, see: https://tox.readthedocs.io/en/latest/config.html#generative-envlist
 
 [tox]
 envlist =
     check
-    {py27,py34,py35,py36,py37,pypy,pypy3}-{t310,t43,t44,t45}-{c45}
+    {py27,py34,py35,py36,py37,pypy,pypy3}-{t310-x27,t43-x27,t44-x28,t45-x28}-{c45}
+    py37-t310-x22-c45
```
blueyed: Add a comment that this is the minimal supported config?  (if that's the case)

tox.ini
```diff
@@ -19,8 +20,9 @@ setenv =
     t44:  _DEP_PYTEST=pytest==4.4.2
     t45:  _DEP_PYTEST=pytest==4.5.0
 
-    {t310,t40,t41,t43}: _DEP_PYTESTXDIST=pytest-xdist==1.27.0
-    {t44,t45}: _DEP_PYTESTXDIST=pytest-xdist==1.28.0
+    x22: _DEP_PYTESTXDIST=pytest-xdist==1.22.0
+    x27: _DEP_PYTESTXDIST=pytest-xdist==1.27.0
+    x28: _DEP_PYTESTXDIST=pytest-xdist==1.28.0
```
blueyed: That looks nice/better in general.
How about using more explicit names here, i.e. "xdist122" etc?  (or "xdist1.22" if that is supported by tox)

graingert: I was continuing with the style already here. Do you want me to expand "t" to "pytest" too?

tox.ini
graingert: It's not minimum, just an arbitrary xdist 1.22.0 test environ

docs/xdist.rst
graingert: I think this change is out of scope

blueyed: ?
It is ok to fix typos / documentation when touching it anyway..

You / your PR made me look at that and I've seen it, so it should be included I think - there's certainly not a reason to create a new PR for that (especially given the number of jobs it would run ;))

tox.ini
blueyed: Ok. Then it should use the minimum versions maybe?

tox.ini
blueyed: Yeah, I know.
I prefer longer names - and would also welcome it with other factors.

@ionelmc ?

ci/templates/appveyor.yml
```diff
@@ -22,6 +22,9 @@ install:
   - echo PyPy installed
   - pypy --version
 
+  # Upgrade virtualenv for e.g. more-itertools to be handled properly.
+  # Pin it to work around https://github.com/tox-dev/tox/issues/1389.
+  - C:\Python35\python -m pip install -U virtualenv==16.5.0
```
gaborbernat: fix going out under https://dev.azure.com/pypa/virtualenv/_build/results?buildId=11337

tox.ini
graingert: Like py27? That's going out of support sooner than xdist==1.22.0

ci/templates/appveyor.yml
gaborbernat: https://pypi.org/project/virtualenv/16.7.3/

graingert:
> I think updating the code to the `pytest-xdist>1.22.0` without breaking compatibility with `pytest-xdist<=1.22.0` will be too difficult. And convolute the code somewhat.
> 
> If we required `pytest>=4.4` we could be sure that xdist users will have at least `pytest-xdist>=1.28.0` installed.
> 
> Until then I believe testing with xdist==1.22.0 sufficiently justifies using the old terminology for compatibility reasons.

actually it turns out to be really easy to make nice looking compatible code with the new terminology

tox.ini
graingert: I think this is now clear with the longer toxenv factor names

ci/templates/appveyor.yml
blueyed: Thanks, @gaborbernat.
@graingert 
I think we can remove this then.

graingert: @blueyed @gaborbernat I've removed this from the template and the generated appveyor.yml

graingert: ok that broke it - I think it would be best to go back to what I had before

blueyed: An upgrade in general might be needed.
(there's no link to a build log so I cannot really tell)

appveyor.yml
```diff
@@ -24,8 +24,8 @@ install:
   - pypy --version
 
   # Upgrade virtualenv for e.g. more-itertools to be handled properly.
-  # Pin it to work around https://github.com/tox-dev/tox/issues/1389.
-  - C:\Python35\python -m pip install -U virtualenv==16.5.0
+  # Pin it to avoid side-effects.
```
graingert: side-effects?

graingert:
@blueyed I think it would be best to leave the virtualenv changes out of this PR

appveyor.yml
blueyed: bugs.

blueyed:
> @blueyed I think it would be best to leave the virtualenv changes out of this PR

Ok.
Do you want to create a separate PR then?
AFAIK it is needed for CI to pass?

graingert:
It's not needed, https://github.com/pytest-dev/pytest-cov/pull/321/commits/1a382ba9be821e89645324f912cae49fd74938c4 already passed

blueyed:
@graingert 
Sorry, no time/energy for this currently.

Also @ionelmc should approve the change of the factor names anyway.

ionelmc:
The idea with the "t"/"c" abbreviations was that it was consistent with "py". AFAIK there were unabbreviated factors in the past but it seemed out of place given tox default "pyXY" abbreviation. Regardless, it would make really long lines in appveyor.yml.

Is there a particular problem that removing the abbreviations would solve?

graingert:
> The idea with the "t"/"c" abbreviations was that it was consistent with "py". AFAIK there were unabbreviated factors in the past but it seemed out of place given tox default "pyXY" abbreviation. Regardless, it would make really long lines in appveyor.yml.
> 
> Is there a particular problem that removing the abbreviations would solve?

I prefer the lengthened toxenv factor names. It matches the tox documentation factor examples

ionelmc:
Mkay fine. @blueyed this is ready to merge right?

ionelmc:
Ah, almost forgot: there should be an update to AUTHORS/CHANGELOG.rst

graingert:
> Ah, almost forgot: there should be an update to AUTHORS/CHANGELOG.rst

done

PR #325, opened 2019-08-18
minor: improve test_dist_missing_data: doc, assert ret first

blueyed:
Asserting the return value first helps in case it is 120 due to
Coverage.py not being installed correctly.

Ref: https://github.com/pytest-dev/pytest-cov/pull/319#issuecomment-522341601


blueyed:
Canceled CI after a few good builds.

PR #330, opened 2019-08-30
Multiple cov fail under flags Fixes #323

### LINKED ISSUE #323: per source cov-fail-under
so I can run `pytest --cov=/src/redacted/ --cov-fail-under=90 --cov-fail-under=100:/src/redacted/testing/ --cov-fail-under=80:/src/redacted/redacted/`

graingert:
Allows you to run `pytest --cov-fail-under=70 --cov-fail-under=100:test/**`.

Fixes #323.

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -99,14 +99,18 @@ class CovController(object):
             out = '%s %s %s\n' % (s * sep_len, txt, s * (sep_len + sep_extra))
             stream.write(out)
 
-    def summary(self, stream):
+    def summary(self, cov_fail_under, stream):
         """Produce coverage reports."""
-        total = None
+        if self.cov_report:
+            self.__summary(self, stream)
 
-        if not self.cov_report:
+        def total(cfu):
             with _backup(self.cov, "config"):
-                return self.cov.report(show_missing=True, ignore_errors=True, file=_NullFile)
+                return self.cov.report(show_missing=True, ignore_errors=True, include=cfu.include, omit=cfu.omit, file=_NullFile)
 
+        return {cfu: total(cfu) for cfu in cov_fail_under}
+
+    def __summary(self, stream):
         # Output coverage section header.
         if len(self.node_descs) == 1:
             self.sep(stream, '-', 'coverage: %s' % ''.join(self.node_descs))
@@ -133,7 +137,7 @@ class CovController(object):
             skip_covered = isinstance(self.cov_report, dict) and 'skip-covered' in self.cov_report.values()
             options.update({'skip_covered': skip_covered or None})
             with _backup(self.cov, "config"):
-                total = self.cov.report(**options)
+                self.cov.report(**options)
 
         # Produce annotated source code report if wanted.
         if 'annotate' in self.cov_report:
@@ -145,7 +149,7 @@ class CovController(object):
             # Coverage.annotate don't return any total and we need it for --cov-fail-under.
 
             with _backup(self.cov, "config"):
-                total = self.cov.report(ignore_errors=True, file=_NullFile)
+                self.cov.report(ignore_errors=True, file=_NullFile)
             if annotate_dir:
                 stream.write('Coverage annotated source written to dir %s\n' % annotate_dir)
             else:
@@ -155,17 +159,15 @@ class CovController(object):
         if 'html' in self.cov_report:
             output = self.cov_report['html']
             with _backup(self.cov, "config"):
-                total = self.cov.html_report(ignore_errors=True, directory=output)
+                self.cov.html_report(ignore_errors=True, directory=output)
             stream.write('Coverage HTML written to dir %s\n' % (self.cov.config.html_dir if output is None else output))
 
         # Produce xml report if wanted.
         if 'xml' in self.cov_report:
             output = self.cov_report['xml']
             with _backup(self.cov, "config"):
-                total = self.cov.xml_report(ignore_errors=True, outfile=output)
+                self.cov.xml_report(ignore_errors=True, outfile=output)
             stream.write('Coverage XML written to file %s\n' % (self.cov.config.xml_output if output is None else output))
-
-        return total
 
 
 class Central(CovController):
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -44,11 +44,56 @@ def validate_report(arg):
     return values
 
 
-def validate_fail_under(num_str):
-    try:
-        return int(num_str)
-    except ValueError:
-        return float(num_str)
+class FailUnder(object):
+    @staticmethod
+    def __number(num_str):
+        if num_str.endswith('%'):
+            num_str = num_str[:-1]
+        try:
+            return int(num_str)
+        except ValueError:
+            return float(num_str)
+
+    @classmethod
+    def parse(cls, unparsed):
+        items = iter(unparsed.split(":"))
+        limit = cls.__number(next(items, None))
+
+        include = []
+        omit = []
+        for item in items:
+            if item.startswith("-"):
+                omit.append(item[1:])
+                continue
+            if item.startswith("+"):
+                include.append(item[1:])
+                continue
+            include.append(item)
+
+        return cls(
+            limit=limit,
+            include=include or None,
+            omit=omit or None,
+        )
+
+    @classmethod
+    def normalize(cls, unparsed):
+        return str(cls.parse(unparsed))
+
+    def __str__(self):
+        return ":".join(
+            ['{}%'.format(self.limit)]
+            + ['+{}'.format(v) for v in (self.include or [])]
+            + ['-{}'.format(v) for v in (self.omit or [])]
+        )
+
+    def __repr__(self):
+        return "FailUnder.parse({})".format(self)
+
+    def __init__(self, limit, include=None, omit=None):
+        self.limit = limit
+        self.include = include
+        self.omit = omit
 
 
 def validate_context(arg):
@@ -91,8 +136,7 @@ def pytest_addoption(parser):
     group.addoption('--no-cov', action='store_true', default=False,
                     help='Disable coverage report completely (useful for debuggers). '
                          'Default: False')
-    group.addoption('--cov-fail-under', action='store', metavar='MIN',
-                    type=validate_fail_under,
+    group.addoption('--cov-fail-under', nargs="?", action='append', default=[], metavar='MIN', type=FailUnder.normalize,
                     help='Fail if the total coverage is less than MIN.')
     group.addoption('--cov-append', action='store_true', default=False,
                     help='Do not delete coverage but append to current. '
@@ -141,11 +185,13 @@ class CovPlugin(object):
         self.pid = None
         self.cov_controller = None
         self.cov_report = compat.StringIO()
-        self.cov_total = None
+        self.cov_totals = None
         self.failed = False
         self._started = False
         self._disabled = False
         self.options = options
+
+        self._fail_under = [FailUnder.parse(v) for v in getattr(self.options, 'cov_fail_under', [])]
 
         is_dist = (getattr(options, 'numprocesses', False) or
                    getattr(options, 'distload', False) or
@@ -188,8 +234,8 @@ class CovPlugin(object):
         self.cov_controller.start()
         self._started = True
         cov_config = self.cov_controller.cov.config
-        if self.options.cov_fail_under is None and hasattr(cov_config, 'fail_under'):
-            self.options.cov_fail_under = cov_config.fail_under
+        if not self._fail_under and hasattr(cov_config, 'fail_under'):
+            self._fail_under = [FailUnder(limit=cov_config.fail_under)]
 
     def _is_worker(self, session):
         return compat.workerinput(session.config, None) is not None
@@ -237,8 +283,8 @@ class CovPlugin(object):
         return not (self.failed and self.options.no_cov_on_fail)
 
     def _failed_cov_total(self):
-        cov_fail_under = self.options.cov_fail_under
-        return cov_fail_under is not None and self.cov_total < cov_fail_under
+        totals = self.cov_totals
+        return any(totals.get(cfu, 0) < cfu.limit for cfu in self._fail_under)
 
     # we need to wrap pytest_runtestloop. by the time pytest_sessionfinish
     # runs, it's too late to set testsfailed
@@ -257,7 +303,10 @@ class CovPlugin(object):
 
         if not self._is_worker(session) and self._should_report():
             try:
-                self.cov_total = self.cov_controller.summary(self.cov_report)
+                self.cov_totals = self.cov_controller.summary(
+                    self._fail_under,
+                    self.cov_report,
+                )
             except CoverageException as exc:
                 message = 'Failed to generate report: %s\n' % exc
                 session.config.pluginmanager.getplugin("terminalreporter").write(
@@ -266,8 +315,8 @@ class CovPlugin(object):
                     warnings.warn(pytest.PytestWarning(message))
                 else:
                     session.config.warn(code='COV-2', message=message)
-                self.cov_total = 0
-            assert self.cov_total is not None, 'Test coverage should never be `None`'
+                self.cov_totals = {}
+            assert self.cov_totals is not None, 'Test coverage should never be `None`'
             if self._failed_cov_total():
                 # make sure we get the EXIT_TESTSFAILED exit code
                 compat_session.testsfailed += 1
@@ -284,21 +333,24 @@ class CovPlugin(object):
         if self.cov_controller is None:
             return
 
-        if self.cov_total is None:
+        if self.cov_totals is None:
             # we shouldn't report, or report generation failed (error raised above)
             return
 
         terminalreporter.write('\n' + self.cov_report.getvalue() + '\n')
+        for cfu in self._fail_under:
+            if cfu.limit <= 0:
+                continue
 
-        if self.options.cov_fail_under is not None and self.options.cov_fail_under > 0:
-            failed = self.cov_total < self.options.cov_fail_under
+            total = self.cov_totals.get(cfu, 0)
+            failed = total < cfu.limit
             markup = {'red': True, 'bold': True} if failed else {'green': True}
             message = (
-                '{fail}Required test coverage of {required}% {reached}. '
+                '{fail}Required test coverage of {required} {reached}. '
                 'Total coverage: {actual:.2f}%\n'
                 .format(
-                    required=self.options.cov_fail_under,
-                    actual=self.cov_total,
+                    required=cfu,
+                    actual=total,
                     fail="FAIL " if failed else "",
                     reached="not reached" if failed else "reached"
                 )
--- a/tests/test_pytest_cov.py
+++ b/tests/test_pytest_cov.py
@@ -5,6 +5,7 @@ import platform
 import re
 import subprocess
 import sys
+import textwrap
 from itertools import chain
 
 import coverage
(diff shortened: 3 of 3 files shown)


src/pytest_cov/engine.py
```diff
@@ -12,6 +14,19 @@
 from .compat import StringIO, workeroutput, workerinput
 
 
+_CovResult = collections.namedtuple("_CovResult", "result,config")
+
+
+def _work_around_config_mutation(cov, method, *args, **kwargs):
```
nedbat: This would be better as a context manager.

src/pytest_cov/engine.py
```diff
@@ -12,6 +14,19 @@
...
+    fn = getattr(cov, method)
+    old_config = cov.config
+    try:
+        cov.config = config = copy.copy(old_config)
```
nedbat: To be extra safe, you want deepcopy.

src/pytest_cov/engine.py
```diff
@@ -107,29 +127,26 @@ def summary(self, stream):
...
                 stream.write('Coverage annotated source written to dir %s\n' % annotate_dir)
             else:
                 stream.write('Coverage annotated source written next to source\n')
 
         # Produce html report if wanted.
         if 'html' in self.cov_report:
-            total = self.cov.html_report(ignore_errors=True, directory=self.cov_report['html'])
-            stream.write('Coverage HTML written to dir %s\n' % self.cov.config.html_dir)
+            v = _work_around_config_mutation(self.cov, "html_report", ignore_errors=True, directory=self.cov_report['html'])
+            stream.write('Coverage HTML written to dir %s\n' % v.config.html_dir)
```
nedbat: This sets v to the return from html_report, which is a float, then uses v as if it were cov.

src/pytest_cov/engine.py
```diff
@@ -107,29 +127,26 @@ def summary(self, stream):
...
 
         # Produce xml report if wanted.
         if 'xml' in self.cov_report:
-            total = self.cov.xml_report(ignore_errors=True, outfile=self.cov_report['xml'])
-            stream.write('Coverage XML written to file %s\n' % self.cov.config.xml_output)
+            v = _work_around_config_mutation(self.cov, "xml_report", ignore_errors=True, outfile=self.cov_report['xml'])
+            stream.write('Coverage XML written to file %s\n' % v.config.xml_output)
```
nedbat: Again.

src/pytest_cov/plugin.py
```diff
@@ -47,6 +47,54 @@ def __call__(self, parser, namespace, values, option_string=None):
         namespace.cov_report[report_type] = file
 
 
+class FailUnder(object):
```
nedbat: I don't understand what this class is for, or does.  Docstrings?

src/pytest_cov/engine.py
graingert: It won't work as a contextmanager

src/pytest_cov/engine.py
graingert: Fortunately coverage only mutates self.config.__dict__

src/pytest_cov/engine.py
graingert: v is a _CovResult

src/pytest_cov/plugin.py
graingert: This is to abstract the limit, include and exclude for each `--cov-fail-under`

src/pytest_cov/engine.py
nedbat: I see: this function also wraps the return in a _CovResult.  This is hard to follow.

nedbat: I would move the _CovResult aspect out of this function, and make it a context manager.

graingert: This is so there is pure function interface to the cov API. A context manager wouldn't provide that

graingert: It's also only a temporary workaround, I'll follow up with a PR to coveragepy

nedbat: So this will only work for coverage.py 5.0+?

graingert: The workaround would version dependent if we wanted to support old coveragepy versions

src/pytest_cov/engine.py
nedbat: You don't like this?
```
@contextlib.contextmanager
def save_cov_config(cov):
    old_config = cov.config
    cov.config = copy.copy(old_config)
    try:
        yield
    finally:
        cov.config = old_config
```
Then you can write:
```
with save_cov_config(cov):
    total = self.cov.html_report(ignore_errors=True, directory=self.cov_report['html'])
res = _CovResult(total, cov.config)
```
This is easier to understand and maintain. If you want a wrapper for _CovResult, you can write that too.

graingert: I prefer a simple function

nedbat: I wish we had a simple one. The callers have to twist themselves, passing the name of a method to call.  The wrapping in _CovResult isn't even hinted at in the name of the function.
My context manager at least only does one thing.

Most of all, I wish we knew who the maintainer of this plugin was.

src/pytest_cov/engine.py
ionelmc: It would work as a context manager, eg:
```py
@contextmanager
def restore(obj, attr):
    backup = getattr(obj, attr)
    try:
        yield
    finally:
        setattr(obj, attr, backup)
```


I think using `with restore(self.cov, 'config'):` would be more clear to the reader about what is actually going on than `_work_around_config_mutation(lots of arguments)`.

src/pytest_cov/engine.py
ionelmc: Hey, I maintain it, or at least I try :/


Using a context manger is more orthogonal and extensible as it can wrap multiple cov method calls or whatever.

src/pytest_cov/engine.py
graingert: ok that context manager looks nice @ionelmc

tests/test_pytest_cov.py
```diff
@@ -357,6 +358,56 @@ def test_cov_min_float_value(testdir):
...
+                               '--cov-fail-under=100.0:+test/*',
+                               '--cov-fail-under=33.33:-test/*',
+                               testsubdir)
+    assert result.ret == 0
+    result.stdout.fnmatch_lines([
+        "Required test coverage of 55.55% reached. Total coverage: 55.56%",
+        "Required test coverage of 100.0%:+test/* reached. Total coverage: 100.00%",
+        "Required test coverage of 100.0%:+test/* reached. Total coverage: 100.00%",
+        "Required test coverage of 33.33%:-test/* reached. Total coverage: 33.33%",
+    ])
```
ionelmc: So to understand what the argument do - in this test it's like this right?
- overall coverage must be at least 55.55%
- test coverage must be 100%
- test coverage must be 100% (alternative syntax with "+")
- overall coverage but without tests must be at least 33.33%

Also if I understand it correctly "%" is optional, and "+" as well yes?

And `--cov-fail-under=100:foo:bar` means coverage must be 100% for `foo` **and** `bar` right? It should be part of the test suite.

There are few things that I think users will find confusing or produce undesired results:
- the multiple syntax to obtain the same thing
- the reporting discrepancy - you can set fail-under for something that you cannot see in the reporting - suddenly you can't know what to fix to make the suite pass

Do we really need the omit mode (the "-" syntax)?

src/pytest_cov/plugin.py
```diff
@@ -41,11 +41,52 @@ def validate_report(arg):
...
+            num_str = num_str[:-1]
+        try:
+            return int(num_str)
+        except ValueError:
+            return float(num_str)
+
+    @classmethod
+    def parse(cls, unparsed):
+        items = iter(unparsed.split(":"))
+        limit = cls.__number(next(items, None))
```
ionelmc: We could use the validate_fail_under from the other merged PR here.

graingert: That's gone now

tests/test_pytest_cov.py
graingert: > So to understand what the argument do - in this test it's like this right?
> 
>     * overall coverage must be at least 55.55%
> 
>     * test coverage must be 100%
> 
>     * test coverage must be 100% (alternative syntax with "+")
> 
>     * overall coverage but without tests must be at least 33.33%
> 
> 
> Also if I understand it correctly "%" is optional, and "+" as well yes?
> 
> And `--cov-fail-under=100:foo:bar` means coverage must be 100% for `foo` **and** `bar` right? It should be part of the test suite.
> 
> There are few things that I think users will find confusing or produce undesired results:
> 
>     * the multiple syntax to obtain the same thing
I don't think it's confusing. You get the same with +1 and 1
> 
>     * the reporting discrepancy - you can set fail-under for something that you cannot see in the reporting - suddenly you can't know what to fix to make the suite pass
I don't believe you can
> 
> 
> Do we really need the omit mode (the "-" syntax)?
Yes so you can get an idea of coverage for code without tests. This is the coverage result people ignore the test suite for.

src/pytest_cov/plugin.py
graingert: It also doesn't need a docstring because it's internal

tests/test_pytest_cov.py
graingert: Most people upgrading to this syntax will have ignored tests globally and will still want to assert on that number

src/pytest_cov/plugin.py
nedbat: Docstrings are helpful for people reading the code to understand what it does, whether it's internal or externally exposed.

tests/test_pytest_cov.py
ionelmc: > Yes so you can get an idea of coverage for code without tests. This is the coverage result people ignore the test suite for.

Most people have just one top level package thus omit is unnecessary.

> I don't believe you can

The reporting show percentages for individual files, doesn't show any totals for packages, subpackages and so on. This is why I don't think this is a great idea.

@nedbat do you have any input on this syntax? I don't want to have something in pytest-cov that will never ever be part of coveragepy.

graingert: > > Yes so you can get an idea of coverage for code without tests. This is the coverage result people ignore the test suite for.
> 
> Most people have just one top level package thus omit is unnecessary
That doesn't follow:
Most people have just one top level package
thus some people do not have just one top level package
thus omit is necessary
> 
> > I don't believe you can
> 
> The reporting show percentages for individual files, doesn't show any totals for packages, subpackages and so on. This is why I don't think this is a great idea.
No the reporting lists a percentage for each --cov-fail-under

> @nedbat do you have any input on this syntax? I don't want to have something in pytest-cov that will never ever be part of coveragepy.
This interface is already available in coveragepy: you just run "coverage report" with different include/omit overrides

mjtorn:
I'm also suffering of `pytest: error: argument --cov-fail-under: invalid int value: '54.20'` style errors.

Sadly this PR (at commit 7989cb6efecaf19fee87f4b3f5c28de010f74db1 at least) breaks `pytest-xdist`:

`INTERNALERROR> DumpError: can't serialize <class 'pytest_cov.plugin.FailUnder'>`

Uninstalling `pytest-xdist` and running a file works just fine:

> FAIL Required test coverage of 54.21% not reached. Total coverage: 54.20%

> Required test coverage of 54.1% reached. Total coverage: 54.20%

Of course in this day and age computers tend to have more than one core, so it would be super nice if `FailUnder` magically became serializable ;)

Thanks for taking on this work, though, much appreciated!

graingert:
@mjtorn do you have more of that stacktrace?

mjtorn:
@graingert sure. Wasn't thinking it's necessary cuz `pip install pytest-xdist` should be enough to break it, but I think the below is good enough without giving any secrets away.

```
============================= test session starts ==============================
platform linux2 -- Python 2.7.15rc1, pytest-4.6.2, py-1.8.0, pluggy-0.12.0 -- REDACTED_VENV/bin/python2
cachedir: .pytest_cache
Django settings: REDACTED_PROJECT.test_settings (from command line option)
rootdir: REDACTED_ROOT, inifile: setup.cfg
plugins: forked-1.0.2, cov-2.7.1, django-3.5.0, xdist-1.28.0
gw0 I / gw1 I / gw2 I / gw3 I / gw4 I / gw5 I / gw6 I / gw7 I

[gw0] linux2 Python 2.7.15 cwd: REDACTED_ROOT
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/_pytest/main.py", line 204, in wrap_session
INTERNALERROR>     config.hook.pytest_sessionstart(session=session)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/pluggy/hooks.py", line 289, in __call__
INTERNALERROR>     return self._hookexec(self, self.get_hookimpls(), kwargs)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/pluggy/manager.py", line 87, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/pluggy/manager.py", line 81, in <lambda>
INTERNALERROR>     firstresult=hook.spec.opts.get("firstresult") if hook.spec else False,
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/pluggy/callers.py", line 208, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/pluggy/callers.py", line 81, in get_result
INTERNALERROR>     _reraise(*ex)  # noqa
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/pluggy/callers.py", line 187, in _multicall
INTERNALERROR>     res = hook_impl.function(*args)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/xdist/dsession.py", line 81, in pytest_sessionstart
INTERNALERROR>     nodes = self.nodemanager.setup_nodes(putevent=self.queue.put)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/xdist/workermanage.py", line 64, in setup_nodes
INTERNALERROR>     nodes.append(self.setup_node(spec, putevent))
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/xdist/workermanage.py", line 73, in setup_node
INTERNALERROR>     node.setup()
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/xdist/workermanage.py", line 254, in setup
INTERNALERROR>     self.channel.send((self.workerinput, args, option_dict, change_sys_path))
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 708, in send
INTERNALERROR>     self.gateway._send(Message.CHANNEL_DATA, self.id, dumps_internal(item))
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1345, in dumps_internal
INTERNALERROR>     return _Serializer().save(obj)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1363, in save
INTERNALERROR>     self._save(obj)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1381, in _save
INTERNALERROR>     dispatch(self, obj)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1466, in save_tuple
INTERNALERROR>     self._save(item)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1381, in _save
INTERNALERROR>     dispatch(self, obj)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1462, in save_dict
INTERNALERROR>     self._write_setitem(key, value)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1456, in _write_setitem
INTERNALERROR>     self._save(value)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1381, in _save
INTERNALERROR>     dispatch(self, obj)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1452, in save_list
INTERNALERROR>     self._write_setitem(i, item)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1456, in _write_setitem
INTERNALERROR>     self._save(value)
INTERNALERROR>   File "REDACTED_VENV/local/lib/python2.7/site-packages/execnet/gateway_base.py", line 1379, in _save
INTERNALERROR>     raise DumpError("can't serialize {}".format(tp))
INTERNALERROR> DumpError: can't serialize <class 'pytest_cov.plugin.FailUnder'>
```

graingert:
@mjtorn nice thanks for that

mjtorn:
You're very welcome. As a workaround, I installed that one commit of yours and run coverage reports without parallelization, which is acceptable for now, but I'll be keeping an eye on this PR, and hopefully a merge and new version will arise :)

graingert:
I think master might fix your issue if you're not using multiple
cov-fail-under options

graingert:
@mjtorn it looks like it's supposed to be fixed https://github.com/pytest-dev/pytest-xdist/issues/384

mjtorn:
Doesn't look like a win, got 1.28.0 here, but maybe someone will react on that issue. Subscribed to its notifications, thanks :)

graingert:
@mjtorn 	try that

mjtorn:
`pip install -U 'git+https://github.com/graingert/pytest-cov@94f5ff3d1ef4bc8fee33a1d3fa25478c332581d0#egg=pytest-cov'` (instead of 7989cb6efecaf19fee87f4b3f5c28de010f74db1)

`pytest -n auto -s -v --ds=REDACTED_PROJECT.test_settings --disable-warnings --cov-report=term-missing --cov-report=xml --cov=. --cov-branch` yielded a reasonable result!

Likewise `pytest -n auto -s -v --ds=REDACTED_PROJECT.test_settings --disable-warnings --cov-report=term-missing --cov-report=xml --cov=. --cov-branch REDACTED/test_REDACTED.py`

--------

The weird thing here is that with both your commit and vanilla `pytest-cov==2.7.1` I get subtly different coverage reports for the same commit of this code base!

I'll be away for quite a few days starting later today, so I don't know how much testing and reporting I can get done. The variations in the coverage reports are quite strange, and I have no proper explanation for them. It could be whoever wrote some of these tests wrote them to leak state that affects the code paths for different `pytest-xdist` workers. In that case it's harmless from a pytest point of view, but definitely something to investigate and fix.

-----------

To sum it up, your latest change feels nice. Are there many blockers for getting it merged? Issues I should be aware of?

nedbat:
As part of my proposal that pytest-cov should do less (#337), I think this shouldn't be merged.  This entire feature can be a separate tool. It's got nothing to do with running tests.

graingert:
@mjtorn does this have the same "subtly different coverage reports" https://github.com/pytest-dev/pytest-cov/pull/338 ?

mjtorn:
@graingert I should be back on this tomorrow, but before going away, I got quite sure the problem was in the tests themselves. I'll let you know, but taking a quick look at #338 I have no idea what I should be looking out for. When do the configs mutate?

@nedbat I'll read through #337 later, but having two severe bugs such as "fail the test if it's less than n.d" failing because `n.d` is not an integer, contrary to Coverage, and a bug where `xdist` don't work, are definitely two things to fix. Anything beyond those bugs, if pytest-cov does too much, I don't care about. Also if this doesn't get merged, I'll have to, what, manually start exiting Jenkins scripts with `1` instead of using a fixed-up previously-broken cli api?-)

mjtorn:
@nedbat I did not write the issues up because it looked like they're being tackled quickly here.

Bug 1: Asking pytest-cov to fail if the coverage is beneath a threshold works only for integers. Quite severe!

Bug 2: xdist didn't work. Blocker-level.

-----

The other fluff about indeterminate results is - with a guesstimated certainty of 99% - someone else's inexperience in writing isolated tests in an existing code base and irrelevant to pytest, coverage or any combination thereof.

graingert:
@mjtorn Bug 1 should be fixed in master already. master is also not impacted by Bug 2

nedbat:
@mjtorn Bug 1 is a great example of why pytest-cov shouldn't be doing all this. Coverage.py has allowed a float for fail-under since version 4.5, 2018-02-03.  Why should we have to also update the UI of the plugin, just so it can pass it along to coverage.py?  If you do your reporting as a separate explicit step using coverage.py directly, then you don't have to wait for two projects to update in order to use the feature.  There's no reason for pytest-cov to be handling this value.

mjtorn:
@graingert you're absolutely right. I got carried away finding everything under the sun except checking the master branch first ;)

@nedbat I'm happy to see that the master branch should work, so all philosophizing is moot. I will say that I disagree with needing an extra step to deal with or work around something that's a part of the documented CLI interface, only because it's buggy in the 2.7.1 release. I don't have time to think about where to draw lines for this stuff, but software should be convenient to use and implementing `--cov-fail-under` as a separate thing is not convenient to use. _If_ that's what The Powers That Be :tm: want to do, I guess I'd be stuck with it. But all of this appears to be moot now.

Best of luck to figuring things out and I'll pop back in if I have the need to. Thanks!

nedbat:
@graingert I'll reiterate my strong belief that it doesn't make sense for pytest-cov to be doing this.  Pytest should run tests and report on whether they pass or fail. Coverage reporting commands can deal with coverage reporting.

graingert:
@nedbat I couldn't disagree more. However I wouldn't object to moving the summary method into the coverage.Coverage class and implementing multiple reporting there

nedbat:
@graingert We should come to an agreement.  This issue (https://github.com/pytest-dev/pytest-cov/issues/337) had overwhelming support for removing reporting features from pytest-cov.  There's no reason for it to implement them.  Whatever people are using to run pytest, they can use that thing to also run coverage reporting commands afterward.  

This just adds needless complexity and awkard interfaces.

PR #334, opened 2019-09-04
Update test deps

ionelmc:
Also test py3.6, 3.7 and pypy3 on Appveyor.


ionelmc:
@graingert regarding your release question: this pr + #304 and then I can make a release.

PR #338, opened 2019-09-12
work around mutating config

graingert:


graingert:
hmm no idea why c34480de1075bb69e7561513f98e6aa88145795d isn't failing on CI. It fails locally

PR #339, opened 2019-09-16
Avoid importing coverage for test runs that don't need it

boxed:
This saves me ~0.1s when running tests without coverage where this plugin is installed.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -37,3 +37,4 @@ Authors
 * Martín Gaitán - https://github.com/mgaitan
 * Hugo van Kemenade - https://github.com/hugovk
 * Michael Manganiello - https://github.com/adamantike
+* Anders Hovmöller - https://github.com/boxed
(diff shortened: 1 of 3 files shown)


blueyed:
Hmm, couldn't src/pytest_cov/engine.py be skipped already?

boxed:
plugin.py imports engine.py, so no :(

blueyed:
Well, that could/should be made conditional there then, no?

boxed:
Yea I guess. I don't have a preference :P

blueyed:
Would result in less conditional imports, i.e. smaller diff I assume.

boxed:
Yea, but less consistent as it'd be conditional includes of two different things...

blueyed:
?

I think altogether it is even better, since it avoids another (not so slow) import then additionally in the first place.

boxed:
Ok, I'll change my PR.

boxed:
Woot! Can we get this merged?

ionelmc:
As soon I get the CI fixed.

PR #342, opened 2019-09-22
--cov-contexts support

nedbat:


ionelmc:
It seems that you have made an uneditable PR - there's an option somewhere that allows maintainers to push to your branch. Closed in favor of #345 .

nedbat:
I think it was "uneditable" because it was in my own fork.

PR #344, opened 2019-09-26
Next release

ionelmc:
So the plan is to upload it on Monday. Comment if I got something wrong in the files.

After this release I guess we start cutting compatibility with old stuff and then we see about #337. 

@nedbat, @graingert


nedbat:
I would love to get #342 merged before that release.

nedbat:
I am very close to merging #345, but I think I need to make a change in coverage.py first.  Can we hold off a day or two?

ionelmc:
We can do anything, that date ain't written in stone. But we can also do a 2.9.0 with just the context stuff - to not feel pressed to make a coveragepy release quickly. What will it be?

nedbat:
I was going to change an API from using glob to using regex.  But I can handwave over that.  Let me write the docs, and then #345 will be ready to merge.

nedbat:
I've changed my mind again. Go ahead and do this release without #345.

nedbat:
Or wait a day or two for me.  Your choice.

nedbat:
#345 is ready to go. Please merge it and then release.

ionelmc:
@nicoddemus any idea why I can't cancel appveyor builds on this project?

ionelmc:
So it looks like this might be broken with xdist 0.30, ooof.

blueyed:
Looks like it might be only due to the header?

> remains unmatched: '*- coverage: platform *, python * -*'

ionelmc:
Workers don't appear to return coverage data. I suspect this is the cause https://github.com/pytest-dev/pytest-xdist/pull/448/files#r330989777

ionelmc:
Mkay so it looks like pytest 5.x and xdist 0.30 break the non-collocated mode (xdist in popen/ssh mode). Pytest 4.6 or xdist 0.29 or earlier work fine. 

Since no one complained yet about releasing delays I'll spend some time investigating this.

ionelmc:
Actually screw it, I'll release it as is, 3.8 is broken as well, and previous pytest-cov release is even more broken.

PR #345, opened 2019-09-28
Dynamic contexts

ionelmc:
This is #342 rebased with a couple changes.

`--cov-context` vs `--cov-contexts` - what's better?

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -1,9 +1,10 @@
 """Coverage plugin for pytest."""
 import argparse
-import coverage
 import os
-import pytest
 import warnings
+
+import coverage
+import pytest
 from coverage.misc import CoverageException
 
 from . import compat
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -1,9 +1,9 @@
 """Coverage plugin for pytest."""
 import argparse
+import coverage
 import os
-import warnings
-
 import pytest
+import warnings
 from coverage.misc import CoverageException
 
 from . import compat
@@ -48,6 +48,14 @@ def validate_fail_under(num_str):
         return float(num_str)
 
 
+def validate_context(arg):
+    if coverage.version_info <= (5, 0):
+        raise argparse.ArgumentTypeError('Contexts are only supported with coverage.py >= 5.x')
+    if arg != "test":
+        raise argparse.ArgumentTypeError('--cov-context=test is the only supported value')
+    return arg
+
+
 class StoreReport(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
         report_type, file = values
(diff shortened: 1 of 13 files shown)


ionelmc:
@nedbat I've added you as a collaborator so you can edit this pr.

nedbat:
I used `--cov-contexts` at first because it will use each test as a context. The result will be measuring many contexts.  But I can see your point.  We can go with `--cov-context`.

src/pytest_cov/plugin.py
```diff
@@ -1,9 +1,9 @@
 """Coverage plugin for pytest."""
 import argparse
+import coverage
```
nedbat: I was sorting imports into sections: stdlib, third-party, local.

src/pytest_cov/plugin.py
```diff
@@ -328,3 +339,33 @@ def cov(request):
...
+def pytest_sessionstart(session):
+    config = session.config
+    cov = config.pluginmanager.get_plugin('_cov')
+    if (
+        config.known_args_namespace.cov_context == 'test'
+        and cov
+        and cov.cov_controller
+        and config.known_args_namespace.cov_source
+    ):
+        config.pluginmanager.register(TestContextPlugin(cov.cov_controller.cov), '_cov_contexts')
```
nedbat: I thought it was easier to add this registration to the existing pytest_sessionstart, since this code is digging into the other plugin's details anyway.

nedbat: Hmm, your comment mentions the old way not working with xdist.  I guess we need to improve the tests then to deal with that case?

ionelmc: Hmmmm ... this could be merged into the main plugin's pytest_sessionstart. Previously it was done in the plugin's `__init__` - that's why it didn't work in xdist mode.

nedbat: My test uses `@xdist_params`, and still passed the old way.  I guess the test isn't strict enough yet?  BTW: IRC?

tox.ini
```diff
@@ -66,9 +67,11 @@ deps =
     flake8
     readme-renderer
     pygments
+    isort
```
graingert: Imho we should copy the pytest core linting   https://github.com/pytest-dev/pytest/blob/master/tox.ini#L52

nedbat: That's a good idea for after this pull request.

nedbat:
@ionelmc This is ready to merge.  Any concerns?

nedbat:
(I've rebased to squash commits to a reasonable set)

ionelmc:
I don't mind many commits, just so you know. Don't have any other comments.

@blueyed have any comments on this PR?

graingert:
@ionelmc how will this handle session and class fixtures? It seems they will be applied to the first test that requires one in each scope.

nedbat:
Yes, the information we get from pytest attributes fixture setup to the first test that needs it, and teardown to the last test that needs it.  This isn't great.  I'd love to know if there are better hooks into pytest to get better information.

PR #347, opened 2019-10-04
Add coverage branch example

mjhea0:


examples/branch/example/__init__.py
```diff
@@ -0,0 +1,17 @@
+
```
ionelmc: Regarding this new directory in examples - I would rather just have the 2 existing examples changed to have the branch options on.

ssbarnea:
Closing this as this was not touched in more than two years, so I doubt it would suddenly become something ready for review. Feel free reopen if you disagree.

Doing this to cleanup the queue, so we can focus on reviewing things that are ready for review and merge.

PR #350, opened 2019-10-05
handle exception, otherwise p.close() will hang

njzjz:
If an exception is raised in f, p.close() and p.join() will hang. We need p.terminate() to handle exception.

### DIFF
--- a/tests/test_pytest_cov.py
+++ b/tests/test_pytest_cov.py
@@ -1233,6 +1233,7 @@ def test_run_target():
         raise
     else:
         # new in python 3.7
+        import sys
         if sys.version_info >= (3, 7):
             p.close()
     finally:
--- a/tests/test_pytest_cov.py
+++ b/tests/test_pytest_cov.py
@@ -1231,9 +1231,10 @@ def test_run_target():
     except:
         p.terminate()
         raise
-    # new in python 3.7
-    #else:
-    #    p.close()
+    else:
+        # new in python 3.7
+        if sys.version_info >= (3, 7):
+            p.close()
     finally:
         p.join()
 ''')
--- a/tests/test_pytest_cov.py
+++ b/tests/test_pytest_cov.py
@@ -1079,8 +1079,12 @@ def test_run_target():
         p = multiprocessing.Pool(3)
         try:
             p.map(target_fn, [i * 3 + j for j in range(3)])
-        finally:
+        except:
             p.terminate()
+            raise
+        else:
+            p.close()
+        finally:
             p.join()
 ''' % ''.join('''if a == %r:
         return a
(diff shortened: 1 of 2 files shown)


blueyed:
Thanks.
Can this be added/covered in the tests?  (not really checked/compared them now)
(`test_multiprocessing_pool_terminate`, `test_multiprocessing_process_with_terminate`)

njzjz:
Is it enough?

tests/test_pytest_cov.py
```diff
@@ -1220,11 +1224,17 @@ def target_fn():
...
-    p.join()
+    try:
+        p.start()
+        time.sleep(0.5)
+        event.wait(1)
+    except:
+        p.terminate()
+        raise
+    else:
+        p.close()
```
ionelmc: There is no Process.close method?

ionelmc:
> Is it enough?

I know we have some broken tests in certain pythons/configurations but new tests need to pass :)

tests/test_pytest_cov.py
njzjz: https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.close
It's just new in Python 3.7.

njzjz: I could comment this line, and enable it when all tests use python>=3.7 (maybe few years later...).

tests/test_pytest_cov.py
```diff
@@ -1231,8 +1231,9 @@ def test_run_target():
...
         p.terminate()
         raise
-    else:
-        p.close()
+    # new in python 3.7
+    #else:
+    #    p.close()
     finally:
         p.join()
 ''')
```
blueyed: Enable/use it with `sys.version_info >= (3, 7)` then?

njzjz: Good point!

njzjz:
> > Is it enough?
> 
> I know we have some broken tests in certain pythons/configurations but new tests need to pass :)

Now CI has passed on two functions that I modified. 😃

tests/test_pytest_cov.py
```diff
@@ -1096,7 +1100,7 @@ def test_run_target():
     assert not testdir.tmpdir.listdir(".coverage.*")
     result.stdout.fnmatch_lines([
         '*- coverage: platform *, python * -*',
-        'test_multiprocessing_pool* 100%*',
+        'test_multiprocessing_pool* 99%*',
```
ionelmc: These assertions shouldn't be changed.

tests/test_pytest_cov.py
```diff
@@ -1234,7 +1246,7 @@ def test_run_target():
 
     result.stdout.fnmatch_lines([
         '*- coverage: platform *, python * -*',
-        'test_multiprocessing_process* 16 * 100%*',
+        'test_multiprocessing_process* 22 * 8*%*',
```
ionelmc: Same here.

tests/test_pytest_cov.py
njzjz: Yes, but I have no idea about how to skip L1082-1084 when calculating the coverage. https://github.com/pytest-dev/pytest-cov/blob/5f0e50398b801d6711480d63669a330d1ba3822a/tests/test_pytest_cov.py#L1082-L1084

PR #354, opened 2019-10-26
Fix comparison against installed PyTest version

adamantike:
With the current code, introduced in #230, the comparison
`pytest.__version__ >= '3.8'` returns the wrong result when using
PyTest 3.10, as `'3.10' < '3.8'` if we use Python's string comparison.

Having `PYTEST_VERSION` as an integer tuple allows us to have simpler
comparisons.

This fixes the following warning when using PyTest 3.10:
```
RemovedInPytest4Warning: config.warn has been deprecated, use warnings.warn instead
```

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -36,3 +36,4 @@ Authors
 * Albert Tugushev - https://github.com/atugushev
 * Martín Gaitán - https://github.com/mgaitan
 * Hugo van Kemenade - https://github.com/hugovk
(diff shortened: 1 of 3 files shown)


ionelmc:
Please also update changelog/authors.rst.

adamantike:
Done, thanks!

adamantike:
@ionelmc, let me know if is there anything else I should add to this PR, to be merged.

ionelmc:
Thanks!

PR #361, opened 2019-11-19
warn specific classes

graingert:


src/pytest_cov/compat.py
```diff
@@ -31,6 +34,21 @@ def testsfailed(self, value):
...
+        return warnings.warn(
+            message=message,
+            category=category,
+            stacklevel=stacklevel+1,
+        )
+else:
+    def warn(config, message, category, stacklevel=1):
+        return config.warn(
+            code=category.code,
+            message=message,
```
graingert: @nicoddemus I'm not sure if this will now break fslocation should I add something like:
```suggestion
            message=message,
            fslocation=_pytest._code.getfslineno(sys._getframe(stacklevel)),
```

graingert: otherwise I could convert this compat functions to return functools.partial objects, and be called like this:

```python
fn = compat.warn(...)
fn.func(*fn.args, **fn.kwargs)
```

ionelmc: The fslocation is sorta useless since it only applies to ancient pytest no?

ionelmc:
So I took the liberty of rebasing - master has the CI fixed now.

src/pytest_cov/plugin.py
```diff
@@ -1,7 +1,6 @@
 """Coverage plugin for pytest."""
 import argparse
 import os
```
graingert, line 3:
```suggestion
import os
import warnings
```

src/pytest_cov/compat.py
```diff
@@ -1,12 +1,15 @@
+import warnings
+
+import pytest
+
```
graingert, line 4:
```suggestion
```

src/pytest_cov/compat.py
```diff
@@ -1,12 +1,15 @@
...
 try:
     from StringIO import StringIO
 except ImportError:
     from io import StringIO
 
-import pytest
-
 StringIO  # pyflakes, this is for re-export
 
+PYTEST_VERSION = tuple(int(v) for v in pytest.__version__.split('.')[:3])
```
graingert, line 8:
```suggestion
import pytest

StringIO  # pyflakes, this is for re-export
```

src/pytest_cov/compat.py
```diff
@@ -7,7 +7,6 @@
 
 StringIO  # pyflakes, this is for re-export
 
-
 if hasattr(pytest, 'hookimpl'):
```
graingert, line 10:
```suggestion

if hasattr(pytest, 'hookimpl'):
```

src/pytest_cov/plugin.py
```diff
@@ -294,7 +311,7 @@ def pytest_terminal_summary(self, terminalreporter):
             if self.options.no_cov_should_warn:
                 message = 'Coverage disabled via --no-cov switch!'
                 terminalreporter.write('WARNING: %s\n' % message, red=True, bold=True)
-                warnings.warn(pytest.PytestWarning(message))
+                warnings.warn(CovReportWarning(message))
```
graingert, line 314:
```suggestion
                warnings.warn(CovDisabledWarning(message))
```

src/pytest_cov/plugin.py
```diff
@@ -14,6 +14,23 @@ class CoverageError(Exception):
...
+
+
+class CovDisabledWarning(PytestCovWarning):
+    """Indicates that Coverage was manually disabled"""
+    code = "COV-1"
+
+
+class CovReportWarning(PytestCovWarning):
+    """Indicates that we failed to generate a report"""
+    code = "COV-2"
```
graingert, line 31:
```suggestion
class PytestCovWarning(pytest.PytestWarning):
    """
    The base for all pytest-cov warnings, never raised directly
    """


class CovDisabledWarning(PytestCovWarning):
    """Indicates that Coverage was manually disabled"""


class CovReportWarning(PytestCovWarning):
    """Indicates that we failed to generate a report"""
```

ionelmc:
Seems simple enough, tests pass, I guess we'll see if we break weird warning filters on the next release.

PR #367, opened 2019-11-24
Travis CI: Test on Python 3.8 final

hugovk:
And don't allow it to fail.

These are currently failing/erroring, as they are on `master`. Those that I checked were doing so in the same way in this PR and on `master` (as expected).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ stages:
 matrix:
   fast_finish: true
   allow_failures:
-  - python: '3.8-dev'
+  - python: '3.8'
 jobs:
   include:
     - stage: lint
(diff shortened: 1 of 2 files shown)


blueyed:
From the top of the file:

> # NOTE: this file is auto-generated via ci/bootstrap.py (ci/templates/.travis.yml).

(might have missed that with other PRs from you before, sorry!)

hugovk:
Well spotted!

The last commit on the template, https://github.com/pytest-dev/pytest-cov/commit/c2aa6d6e9f78af07e509c54fa3220e46cd8e5133, explains why 3.8 jobs are failing:

> Allow 3.8 to fail (pending resolve of bugs.python.org/issue38227).

It's still unresolved, so I've updated `3.8-dev` to `3.8` via the scripts, and left the `allow_failures`.

PR #370, opened 2019-12-05
The context tests were failing on newer coverage

nedbat:
but it was because the test was fragile.  This fixes the test.

### DIFF
--- a/tests/test_pytest_cov.py
+++ b/tests/test_pytest_cov.py
@@ -1934,12 +1934,12 @@ def test_cov_and_no_cov(testdir):
 
 
 def find_labels(text, pattern):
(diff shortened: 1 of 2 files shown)


nedbat:
@ionelmc @blueyed Any ideas about these failing tests?

blueyed:
@nedbat haven't looked at them, but likely https://github.com/pytest-dev/pytest-cov/issues/365.

If this PR fixes some (others) I guess @ionelmc still wants to merge it already.
(IMHO we should try making CI green again first, to ensure new PRs do not cause regressions, but that appears to be not trivial)

ionelmc:
Yep, seems fine.

Issue #386, opened 2020-01-16
Combine issues in tox examples for coverage 5


joshringer:
Re: https://github.com/pytest-dev/pytest-cov/commit/d3daf7608a327c10f8b77de462791078a0596434

I have been looking into this a bit today. It appears from the latest coveragepy source that the only reference to the `[paths]` config—or at least the only reference I could find—is [here](https://github.com/nedbat/coveragepy/blob/coverage-5.0.3/coverage/control.py#L673).
This means that the filepaths are not combined properly somehow when merely doing `--cov-append` (I'm guessing this difference has something to do with the new db format for the _.coverage_ data file).

I have managed a workaround by setting a different data file per environment and then combining during the report step, i.e. for _examples/src-layout/tox.ini_:
```ini
...
[testenv]
commands = pytest --cov {posargs:-vv}
setenv = 
    {py27,py38}: COVERAGE_FILE=.coverage.{envname}

...
[testenv:report]
commands =
    coverage combine
    ...
```

I suspect the "correct" fix here is to have pytest-cov respect the `[run] parallel` flag and add a suffix to the data file as vanilla coverage does. This would at least remove the need for the ugly `setenv` above. It would kinda break old behaviour though, needing to add a `coverage combine` down there, so it would probably have to wait for pytest-cov 3.0...

davidism:
Possibly related to https://github.com/nedbat/coveragepy/issues/713

abn:
Facing the same issue here. It would be great if, as per suggestion, `pytest-cov` respects the `parallel` config value and skips combining the data files if `self.cov.config.parallel` is `True`.

In the current implementation, `coverage combine` implicit. This is not necessarily a desired behaviour especially when running tests across multiple python versions, platforms or machines.

Alternatively, if the old behaviour is to be preserved, it would be great to enable a new option like `--cov-skip-combine` that if enabled will skip combining and replacing the default behaviour in a later release.

ionelmc:
IMO pytest-cov shouldn't even allow users changing the `parallel` config as it critical for measuring subprocess coverage.

Having a `--cov-skip-combine` could be a solution, but ugly for both the changes it would incur in the codebase and the CLI argument polution so I'm looking to hearing your arguments for your usecases. So far they are a complete mystery to me :)

abn:
[...]

PR #387, opened 2020-01-20
#386 Always enable data_suffix for --cov-append

Urth:
This push request should fix https://github.com/nedbat/coveragepy/issues/883 and https://github.com/pytest-dev/pytest-cov/issues/386 for users of pytest-cov.

Setting `data_suffix=True` on the [primary coverage instance](https://github.com/pytest-dev/pytest-cov/blob/master/src/pytest_cov/engine.py#L219-L221) will enable `parallel=True` and ensure the coverage processes do not conflict with each other. Because pytest-cov always combines the coverage result this is safe to always enable.

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -218,6 +218,7 @@ class DistMaster(CovController):
 
         self.cov = coverage.Coverage(source=self.cov_source,
                                      branch=self.cov_branch,
+                                     data_suffix=True,
                                      config_file=self.cov_config)
         self.combining_cov = coverage.Coverage(source=self.cov_source,
                                                branch=self.cov_branch,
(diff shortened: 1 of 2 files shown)


blueyed:
@Urth 
Would be great to have some test for this.

Urth:
I force pushed to correct flake8 error but I think this broke the builds that were still testing the now deleted commit. I fear this may need a manual intervention, sorry.

On the bright side I did add a working test for the patch. I'm not sure how xdist interacts with --cov-append and if that is needed for this new test. Please let me know if that needs to change.

ionelmc:
FYI I'm trying to include this and cut a release next week (https://github.com/pytest-dev/pytest-cov/tree/temp). Ping me if i forget about it again :weary:

ionelmc:
Alright so this is now in master.

Urth:
Thanks, works like a charm :ok_hand:

PR #407, opened 2020-06-01
Support -Werror with --no-cov

Zac-HD:
We still print a warning line to the console if the `--no-cov` option was passed, but skipping the Python warning system allows use of `-Werror` mode as well.

Looking back through the git history it doesn't look like there was a strong reason to emit the warning in addition to printing a line to the console, so I've just removed it.

This is particularly useful for projects (like most of mine) which enable both warnings-as-errors and require 100% coverage as pytest defaults; in the current version I can't actually use `--no-cov` at all and that's rather unfortunate when I want to experiment with mutation testing.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,3 +38,4 @@ Authors
 * Hugo van Kemenade - https://github.com/hugovk
 * Michael Manganiello - https://github.com/adamantike
 * Anders Hovmöller - https://github.com/boxed
+* Zac Hatfield-Dodds - https://zhd.dev
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+Future release
+--------------
+
+* ``--no-cov`` can now be use with ``-Werror``.  Contributed by Zac Hatfield-Dodds.
+
+
 2.9.0 (2020-05-22)
 ------------------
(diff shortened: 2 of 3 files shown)


ionelmc:
I don't get why the failed result is a problem in your situation. Does `-Werror` discard certain output that you need? You're supposed to only use `--no-cov` locally - I'm assuming that you don't use that option in CI ...

Zac-HD:
> I don't get why the failed result is a problem in your situation.

I'm using 'mutation testing' tools, which search for ways to mutate (i.e. edit and break) your source code, but for which all your tests pass.  I therefore want to run `pytest -x --no-cov -n auto` on each candidate to reach the first failure as fast as possible, and then bail out - coverage is a ~30% slowdown for me, which is usually fine (~15s) but in this case adds many hours.

A failed result due to `-Werror --no-cov` would incorrectly mark the mutant as detected, whereas in fact my tests just fail regardless unless I edit config files.

ionelmc:
Ok, now it makes sense. I still think the warning should be there to prevent people leaving --no-cov in addopts. See https://github.com/pytest-dev/pytest-cov/pull/408 for what I think is the correct fix.

Zac-HD:
That will definitely work for me, and looks like a better patch all round!  Thanks @ionelmc :smile:

PR #408, opened 2020-06-01
Implement better --no-cov warning

ionelmc:
Only do it if --no-cov is present before --cov. Also remove some legacy pytest support and bump min version in setup.py. Closes #407. Ref #304.


Zac-HD:
Thanks @ionelmc!  I know I'll find this really useful :grin:

PR #410, opened 2020-06-03
Added toml as extra

Cielquan:
For using `coverage` with `pyproject.toml` you need to install `coverage[toml]`. 

For convenience and to reduce the dependency declaration lines for projects using `pytest-cov` and `pyproject.toml` I added the `toml` extra to `pytest-cov` as well.

Before (poetry examples):
```toml
[tool.poetry.dev-dependencies]
pytest-cov = "^2.8.1"
toml = "^0.10.1"
```

Or:
```toml
[tool.poetry.dev-dependencies]
pytest-cov = "^2.8.1"
coverage = {version = "^5", extras = ["toml"]}
```

Now:
```toml
[tool.poetry.dev-dependencies]
pytest-cov = {version = "^2.8.1", extras = ["toml"]}
```


setup.py
```diff
@@ -134,7 +134,8 @@ def run(self):
             'six',
             'pytest-xdist',
             'virtualenv',
-        ]
+        ],
+        'toml': ['toml'],
```
graingert, line 138: shouldn't this be
```suggestion
        'toml': ['coverage[toml] >= WHEN_COVERAGE_GOT_TOML_SUPPORT'],
```

graingert: this way if coverage changes their toml parser, it will still work

Cielquan: I just got the same idea, and tested it locally. But it does not work for me.

This is what I get:
```
$ pip install .[toml]
Processing /home/krys/Python_Projects/.contributions/pytest-cov
Collecting coverage>=4.4
  Using cached coverage-5.1-cp36-cp36m-manylinux1_x86_64.whl (227 kB)
Collecting pytest>=3.6
  Using cached pytest-5.4.3-py3-none-any.whl (248 kB)
Collecting py>=1.5.0
  Using cached py-1.8.1-py2.py3-none-any.whl (83 kB)
Collecting wcwidth
  Using cached wcwidth-0.2.3-py2.py3-none-any.whl (30 kB)
Collecting more-itertools>=4.0.0
  Using cached more_itertools-8.3.0-py3-none-any.whl (44 kB)
Collecting attrs>=17.4.0
  Using cached attrs-19.3.0-py2.py3-none-any.whl (39 kB)
Collecting pluggy<1.0,>=0.12
  Using cached pluggy-0.13.1-py2.py3-none-any.whl (18 kB)
Collecting packaging
  Using cached packaging-20.4-py2.py3-none-any.whl (37 kB)
Collecting importlib-metadata>=0.12; python_version < "3.8"
  Using cached importlib_metadata-1.6.0-py2.py3-none-any.whl (30 kB)
Collecting pyparsing>=2.0.2
  Using cached pyparsing-2.4.7-py2.py3-none-any.whl (67 kB)
Collecting six
  Using cached six-1.15.0-py2.py3-none-any.whl (10 kB)
Collecting zipp>=0.5
  Using cached zipp-3.1.0-py3-none-any.whl (4.9 kB)
Using legacy setup.py install for pytest-cov, since package 'wheel' is not installed.
Installing collected packages: coverage, py, wcwidth, more-itertools, attrs, zipp, importlib-metadata, pluggy, pyparsing, six, packaging, pytest, pytest-cov
    Running setup.py install for pytest-cov ... done
Successfully installed attrs-19.3.0 coverage-5.1 importlib-metadata-1.6.0 more-itertools-8.3.0 packaging-20.4 pluggy-0.13.1 py-1.8.1 pyparsing-2.4.7 pytest-5.4.3 pytest-cov-2.9.0 six-1.15.0 wcwidth-0.2.3 zipp-3.1.0
```

No `toml` .. but when I comment out `'coverage>=4.4'` in the `install_requires` section it does as we want: 

```
$ pip install .[toml]
Processing /home/krys/Python_Projects/.contributions/pytest-cov
Collecting pytest>=3.6
  Using cached pytest-5.4.3-py3-none-any.whl (248 kB)
Collecting coverage[toml]>=5
  Using cached coverage-5.1-cp36-cp36m-manylinux1_x86_64.whl (227 kB)
Collecting packaging
  Using cached packaging-20.4-py2.py3-none-any.whl (37 kB)
Collecting pluggy<1.0,>=0.12
  Using cached pluggy-0.13.1-py2.py3-none-any.whl (18 kB)
Collecting attrs>=17.4.0
  Using cached attrs-19.3.0-py2.py3-none-any.whl (39 kB)
Collecting more-itertools>=4.0.0
  Using cached more_itertools-8.3.0-py3-none-any.whl (44 kB)
Collecting py>=1.5.0
  Using cached py-1.8.1-py2.py3-none-any.whl (83 kB)
Collecting importlib-metadata>=0.12; python_version < "3.8"
  Using cached importlib_metadata-1.6.0-py2.py3-none-any.whl (30 kB)
Collecting wcwidth
  Using cached wcwidth-0.2.3-py2.py3-none-any.whl (30 kB)
Collecting toml; extra == "toml"
  Using cached toml-0.10.1-py2.py3-none-any.whl (19 kB)
Collecting six
  Using cached six-1.15.0-py2.py3-none-any.whl (10 kB)
Collecting pyparsing>=2.0.2
  Using cached pyparsing-2.4.7-py2.py3-none-any.whl (67 kB)
Collecting zipp>=0.5
  Using cached zipp-3.1.0-py3-none-any.whl (4.9 kB)
Using legacy setup.py install for pytest-cov, since package 'wheel' is not installed.
Installing collected packages: six, pyparsing, packaging, zipp, importlib-metadata, pluggy, attrs, more-itertools, py, wcwidth, pytest, toml, coverage, pytest-cov
    Running setup.py install for pytest-cov ... done
Successfully installed attrs-19.3.0 coverage-5.1 importlib-metadata-1.6.0 more-itertools-8.3.0 packaging-20.4 pluggy-0.13.1 py-1.8.1 pyparsing-2.4.7 pytest-5.4.3 pytest-cov-2.9.0 six-1.15.0 toml-0.10.1 wcwidth-0.2.3 zipp-3.1.0
```

I think `coverage[toml]` is completely skipped because `coverage` itself is a install dependency.

Cielquan: I tried to achieve  this with `poetry` too .. but it does not work also.

Cielquan: @graingert After some thinking, googling and testing I think it is impossible to achieve.

graingert: can you post the links you found?

ionelmc:
Oooof. Well, here goes ...

```
[tool.poetry.dev-dependencies]
pytest-cov = {version = "^2.8.1", extras = ["toml"]}
```

If anyone would ask me I'd say that's a bad practice. An antipattern. Why should we even allow users use this footgun? We don't have a upper constraint on coverage because it would create lots of churn thus users will always be at risk of getting a broken coverage+pytest-cov combination in the future (eg: new coverage release). Thus it's always better to specify the coverage version you want in your project settings.

setup.py
graingert: @Cielquan https://github.com/pypa/pip/issues/4957

Cielquan: I found issues on stackoverflow .. dont have the links anymore.

But the pip issue you found is good.

Cielquan:
> Oooof. Well, here goes ...
> 
> ```
> [tool.poetry.dev-dependencies]
> pytest-cov = {version = "^2.8.1", extras = ["toml"]}
> ```
> 
> If anyone would ask me I'd say that's a bad practice. An antipattern. Why should we even allow users use this footgun? We don't have a upper constraint on coverage because it would create lots of churn thus users will always be at risk of getting a broken coverage+pytest-cov combination in the future (eg: new coverage release). Thus it's always better to specify the coverage version you want in your project settings.

I understand your concern and I think you are right for my initial proposal.

With the [change proposed](https://github.com/pytest-dev/pytest-cov/pull/410#discussion_r434436834) by @graingert I think this would fix the issue **IF** the [pip issue](https://github.com/pytest-dev/pytest-cov/pull/410#discussion_r434549839) mention above gets solved.

So there would still be no upper constraint either way and when the user wants to use `pyproject.toml` he/she can use the `toml` extra which is just a 'link' to `coverage`'s `toml` extra. So even if `coverage` changes the extra to another lib it would work. It would also work when `coverage` moves toml to `install_requires` and removes the extra. Pip then gives a warning that the `toml` extra does not exist.

hukkinj1:
> Oooof. Well, here goes ...
> 
> ```
> [tool.poetry.dev-dependencies]
> pytest-cov = {version = "^2.8.1", extras = ["toml"]}
> ```
> 
> If anyone would ask me I'd say that's a bad practice. An antipattern. Why should we even allow users use this footgun? We don't have a upper constraint on coverage because it would create lots of churn thus users will always be at risk of getting a broken coverage+pytest-cov combination in the future (eg: new coverage release). Thus it's always better to specify the coverage version you want in your project settings.

@ionelmc I'm not sure I agree with this. Isn't this what we have lockfiles for (in the case of Poetry, a `poetry.lock`)? That is, so we don't have to manually list and version lock each and every dependency of a dependency (recursively) in `[tool.poetry.dev-dependencies]`? The tooling does it for us.

@Cielquan Pip got it's dependency resolver in v20.3 and the issue you mentioned is closed :partying_face: Maybe this would be doable now?

setup.py
Cielquan: As mentioned by @hukkinj1 the above mentioned pip issue is closed. I tested it again with a fresh venv and pip 20.3 and pip 21.0.1. Both worked just fine.

Cielquan:
I accepted @graingert 's proposal for the chained toml dependency to coverage's one. I tested beforehand again and it worked with a fresh venv and pip 20.3 and pip 21.0.1.

Therefore I make it ready for review.

setup.py
```diff
@@ -135,7 +135,8 @@ def run(self):
             'six',
             'pytest-xdist',
             'virtualenv',
-        ]
+        ],
+        'toml': ['coverage[toml] >= 5'],
```
ionelmc, line 139: Isn't the `>=5` part redundant? It's already specified more precisely in `install_requires`.

graingert:
fyi, pytest depends on toml directly

https://github.com/pytest-dev/pytest/blob/main/setup.cfg#L49

hukkinj1:
> fyi, pytest depends on toml directly
> 
> https://github.com/pytest-dev/pytest/blob/main/setup.cfg#L49

:+1: 
I think it would be good for `pytest-cov` to depend on `coverage[toml]` directly (not via an extra) because
- `toml` is required anyways because of `pytest` so this doesnt add extra dependencies
- if `coverage` starts to restrict `toml` versions, the will the restriction will also apply to `pytest-cov`

ionelmc:
I think we can depend on toml if it doesn't bring in any c extensions or other crazy stuff, just to be practical - I mean that's the idea of pytest-cov, to streamline and remove any warts if possible.

graingert:
>  I think it would be good for pytest-cov to depend on coverage[toml] directly (not via an extra) because
:+1: I believe pip only warns if you depend on a version of coverage without the `toml` extra

hukkinj1:
@graingert Yup correct. And because we require `coverage>=5.2.1` and that version already has the `toml` extra, there is currently no compatible combination of package versions that would even issue that warning.

setup.py
Cielquan: Good point. I don't know.

Cielquan:
1) I think that `pytest-cov` depending on `toml` directly even though it is not a needed dependency of `pytest-cov` itself is not a good idea. Reason is simple. What if `coverage` changes from `toml` to e.g. `tomlkit` (which is IMO better) then we would be at the current point without this PR.

2) Even so `pytest` depends on `toml` directly who says that it will not change to another lib? (similar to 1. point)

3) I also think it is not good practise to depend on `coverage[toml]` from the beginning. When point 1 would come true and point 2 not, which would result in `pytest` depending on `toml` and `coverage` depending on anther toml lib, then someone who does not use `pyproject.toml` would have a _useless_ dependency installed. I would rather highlight the available `toml` extra in the Readme/docs.

Those points have some ifs and point 2 and 3 are not that heavy impactful but those are just my two cents.

hukkinj1:
I'd just add the `install_requires=['coverage[toml]>=5.2.1']`

> When point 1 would come true and point 2 not, which would result in pytest depending on toml and coverage depending on anther toml lib, then someone who does not use pyproject.toml

Obviously we can't predict the future and prepare for everything, given infinite time, but if things change, we can change the dependencies.

As @ionelmc said:
> I mean that's the idea of pytest-cov, to streamline and remove any warts if possible.

Remembering that you need an extra just because you want to have configuration in pyproject.toml to me sounds like one of these warts that pytest-cov could and should remove.

toml is a tiny (~1k LOC) pure python package that due to its popularity is probably installed anyways (i mean quite literally so due to pytest). The downside of it being installed and unused is a) minimal and b) theoretical (due to pytest).

ionelmc:
> I'd just add the `install_requires=['coverage[toml]>=5.2.1']`

That seems fine to me.

Cielquan:
> > I'd just add the `install_requires=['coverage[toml]>=5.2.1']`
> 
> That seems fine to me.

Updated PR accordingly.

PR #412, opened 2020-06-14
Use "worker" for compatibility with xdist >= 2.0

### LINKED ISSUE #411: Upcoming incompatibility in pytest-xdist 2.0: "slave" aliases removed
Hi!

As per https://github.com/pytest-dev/pytest-xdist/pull/541, pytest-xdist 2.0 will remove the deprecated "slave" 
aliases, which have been replaced by "worker" since pytest-xdist 1.22.

To avoid breakage, the team suggests to update your package to use the new terms, and make a new release to avoid sudden breakages once pytest-xdist 2.0 is released.

cc @Zac-HD @RonnyPfannschmidt

Thanks!

Zac-HD:
Fixes #411, at cost of support for xdist 1.22 and earlier from 2017.  Given that such old versions aren't even in our test matrix, I figured that this would be OK.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.10.1 (2020-06-??)
+-------------------
+
+* Support for ``pytest-xdist`` 2.0, which breaks compatibility with ``pytest-xdist`` before 1.22.3 (from 2017).
+  Contributed by Zac Hatfield-Dodds in `#412 <https://github.com/pytest-dev/pytest-cov/pull/412>`_.
+
 2.10.0 (2020-06-12)
 -------------------
 
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -38,3 +38,4 @@ Authors
 * Hugo van Kemenade - https://github.com/hugovk
 * Michael Manganiello - https://github.com/adamantike
 * Anders Hovmöller - https://github.com/boxed
+* Zac Hatfield-Dodds - https://zhd.dev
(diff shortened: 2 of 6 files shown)


ionelmc:
Please also add something in changelog/authors.rst.

Zac-HD:
@ionelmc - done.  Not sure what's happening with the CI failure, but it looks like a sqlite error rather than something related to this PR?

ionelmc:
AppVeyor runs a broken pypy. I've been trying to get pypy up to date in choco but it's slow - https://github.com/refi64/chocolatey-pypy/issues/18

CHANGELOG.rst
```diff
@@ -1,6 +1,12 @@
 Changelog
 =========
 
+2.10.1 (2020-06-??)
```
graingert, line 4: imho this should be a pytest-cov 3 release

Zac-HD: Because it removes some `compat.xxx` names?

If compatibility with pytest-xdist <= 1.22 (from 2018) is really that important - and note that the oldest version pytest-cov is tested against is xdist 1.27 (from 2019) - I could add them back.  @ionelmc?

graingert: there's no harm in bumping the major version, but we don't have to make the decision in this PR

ionelmc: We certainly don't want to extend the testing grid, it's way too big already.

hugovk: As expected, old releases like pytest-xdist 1.22 account for very few downloads:

image

[pepy.tech link](https://pepy.tech/project/pytest-xdist?versions=*&versions=1.32.*&versions=1.31.*&versions=1.30.*&versions=1.29.*&versions=1.28.*&versions=1.27.*&versions=1.26.*&versions=1.25.*&versions=1.24.*&versions=1.23.*&versions=1.22.*)

PR #429, opened 2020-08-28
Update config.rst wrt ignored .coveragerc settings

PamelaM:
I took a while to figure out why coverage wasn't working correctly for our large project until yesterday, when I read the paragraph that called out the `.coveragerc` options that `pytest-cov` ignores.  

This PR makes that part of the documentation "pop" a bit more.

I considered moving it under "Caveats", but decided that that would mean rewriting "Caveats" to have two separate issues, rather than the one that it currently explains.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -21,9 +21,16 @@ For full details refer to the `coverage config file`_ documentation.
 
 .. _`coverage config file`: https://coverage.readthedocs.io/en/latest/config.html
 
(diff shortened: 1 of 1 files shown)


ionelmc:
So I've edited it a bit to be more accurate wrt the actual implementation. Does it still have the clarity you wanted @PamelaM?

PamelaM:
Thank you.  That's much better.

Issue #416, opened 2020-06-27
Running pytest-cov in parallel


JulienPalard:
# Summary

People are using `tox -p auto` or `tox -p all` more and more since it exists, (and some are simply using `&` in shell scripts).

But it's failing with `pytest-cov` (https://github.com/nedbat/coveragepy/issues/883, https://github.com/pytest-dev/pytest-cov/issues/356, https://github.com/pytest-dev/pytest-cov/issues/237, https://github.I seecom/pytest-dev/pytest-cov/issues/217).

This is because pytest-cov uses a `coverage combine` step which tries to combine all `.coverage.*` files, mixing files from all of the parallels runs. As some are incomplete, this often yield to sqlite errors, but it also sometime just mix the data in strange ways.

A clean fix is to specify a specific coverage file name for each run, so the combine step will search for files with this specific name, avoiding mixing the files.

This can easily be done, for example, in `.tox.ini` by using:
```
setenv =
  COVERAGE_FILE=.coverage.{envname}
```

It make `coverage combine` search for `.coverage.py37.*` for example.

I see two strategies:

Either pytest-cov picks a unique coverage file name per run or pytest-cov documents that when used in parallel one should specify a coverage file name to disambiguate the runs.

Do you have a preference?

JulienPalard:
As I read a few issues, it looks like many people are relying on a `.coverage` being generated, so they can typically upload them somewhere, which leaves an open question to me:

Imagine one have a `tox.ini` like:
```ini
[tox]
envlist = py37, py38

[testenv]
deps = -rrequirements.txt
commands = pytest --cov-report term-missing --cov {envsitepackagesdir}/foo/
```

and some Python code like:

```python
if sys.version_info >= (3, 8):
    ...
else:
    ...
```

I don't think we want two distinct coverage files, one whining about the first branch not being tested and the other whining about the other branch not being tested.

I think we need a single `.coverage` file telling that all branches are tested, a "final combination", which, I think, can't be obtained right now as far as I understand.

Using my proposed 
```setenv =
  COVERAGE_FILE=.coverage.{envname}
```
gives 3 output files, `.coverage.py37`, `.coverage.py38`, `.coverage.py39`, which can be combined manually with `coverage combine`, which correctly report 100% coverage.

I'm trying with this "reproducer" https://mdk.fr/x/pytest-cov-reproducer.tar.bz2

[...]

PR #435, opened 2020-09-23
Change example configs to support tox -e pyXX

gschaffner:
`tox -e pyXX` does not run the envs that pyXX depends on [[src](https://tox.readthedocs.io/en/latest/config.html?highlight=depends#conf-depends)]. This means that with the current example configs, running `tox -e pyXX` will not run `coverage clean` and so `pytest-cov` data will be appended to the preexisting `.coverage` file.

IMO, appending to the existing coverage data is not the expected behavior when running a single test suite against one Python version.

In fact, the current behavior can even cause decreased coverage to go undetected. E.g.:
1.  The user runs `tox` and generates a `.coverage` file that reports 100% coverage.
1.  The user deletes a bunch of tests and then runs `tox -e py38`.

With the current example configs, `tox -e py38` will report 100% coverage despite coverage really being lower. The user would have to run `tox` in full to discover the decreased coverage.

Related: #416

### DIFF
--- a/examples/adhoc-layout/tox.ini
+++ b/examples/adhoc-layout/tox.ini
@@ -1,12 +1,14 @@
 [tox]
-envlist = clean,py27,py38,report
+envlist = py27,py38,report
 
 [tool:pytest]
 addopts =
     --cov-report=term-missing
 
 [testenv]
-commands = pytest --cov --cov-append --cov-config={toxinidir}/.coveragerc {posargs:-vv}
+setenv =
+    py{27,38}: COVERAGE_FILE = .coverage.{envname}
+commands = pytest --cov --cov-config={toxinidir}/.coveragerc {posargs:-vv}
 deps =
     pytest
     coverage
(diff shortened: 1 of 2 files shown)


ionelmc:
To be honest I don't like this solution. Perhaps pytest-cov should just handle this particular scenario by default instead of asking users to make arcane configuration (it's already arcane enough as it is). Eg: do tox env name suffixing by default.

gschaffner:
I agree. However, I imagine that appending a suffix to the output `.coverage` file would break some people's tox setups. Probably very few would be affected though, since `coverage combine` searches `.coverage.*` itself. I imagine that most people using `coverage combine` allow it to search rather than specifying data files explicitly.

ionelmc:
Ok so I'm thinking now that maybe making pytest-cov override the coverage file name would me more trouble than it's worth (like breaking existing configurations and usecases). Maybe we need to have your changes in the examples after all ...

Any particular reason the cleanup env was removed?

gschaffner:
I think that the cleanup env is only needed when `--cov-append` is being used. Without `--cov-append`, the coverage files just get overwritten by pytest-cov, so pre-pytest cleanup isn't needed.

PR #438, opened 2020-10-02
Document coverage envvars

nils-werner:
After digging around in the configuration settings of `pytest-cov` and `coverage.py`, looking for a way to configure the output data filename, I came across the environment variable `COVERAGE_FILE`.

This variable is even being used in the tests of `pytest-cov`, but its use is never mentioned.

This PR does two things

 - It normalizes the whitespace in `config.rst` (it is the only file in the docs with superfluous whitespace)
 - It mentions the environment variable, so that future folks don't have to search for it like I did.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -24,7 +24,8 @@ For full details refer to the `coverage config file`_ documentation.
 .. note:: Important Note
 
     This plugin overrides the ``data_file`` and ``parallel`` options of coverage. Unless you also run coverage without
-    pytest-cov it's pointless to set those options in your ``.coveragerc``.
+    pytest-cov it's pointless to set those options in your ``.coveragerc``.  Instead, to override the ``data_file`` setting,
+    you may instead use the environment variable ``COVERAGE_FILE``.
 
     If you use the ``--cov=something`` option (with a value) then coverage's ``source`` option will also get overriden.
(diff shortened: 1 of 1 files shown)


blueyed:
`COVERAGE_FILE` is documented with Coverage.py: https://coverage.readthedocs.io/en/latest/cmd.html?highlight=coverage_file#data-file
I think it would be best to have its envvars documented there (in its docs), and then link to it from pytest-cov maybe.

nils-werner:
Yes, that's where I found it. However currently the [docs in `pytest-cov`](https://pytest-cov.readthedocs.io/en/latest/config.html) are worded in a way that makes it sound like there is no way at all to change this.

To mitigate this confusion, I think it might be useful to add the single sentence that I am proposing.

ionelmc:
I think I need to completely rewrite that section (or remove it), not just add more text/confusion. Cause the plugin doesn't override those options. Maybe it did in the past. Hmmmmm.

ssbarnea:
Closing this as this was not touched in almost two years, so I doubt it would suddenly become something ready for review. Feel free reopen if you disagree.

Doing this to cleanup the queue, so we can focus on reviewing things that are ready for review and merge.

nedbat:
I think in general, it's easy for pytest-cov users to misunderstand that coverage.py is under the hood doing real work, and they are affected by it.  I don't know what to do in the docs to make that clearer, but something could help.

PR #443, opened 2020-11-04
Inherit coverage context in subprocesses

### LINKED ISSUE #442: Support coverage context with subprocess calls
After coverage starts here https://github.com/pytest-dev/pytest-cov/blob/master/src/pytest_cov/embed.py#L74, we should pull context set from parent process https://github.com/pytest-dev/pytest-cov/blob/master/src/pytest_cov/embed.py#L74 here.

gaborbernat:
At the moment subprocess calls are missing coverage context, this fixes that issue.

Resolves #442 

Signed-off-by: Bernát Gábor <bgabor8@bloomberg.net>

### DIFF
--- a/src/pytest_cov/embed.py
+++ b/src/pytest_cov/embed.py
@@ -45,6 +45,7 @@ def init():
     cov_config = os.environ.get('COV_CORE_CONFIG')
     cov_datafile = os.environ.get('COV_CORE_DATAFILE')
     cov_branch = True if os.environ.get('COV_CORE_BRANCH') == 'enabled' else None
+    cov_context = os.environ.get('COV_CORE_CONTEXT')
 
     if cov_datafile:
         if _active_cov:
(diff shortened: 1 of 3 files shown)


ionelmc:
Good catch. Thanks.

ionelmc:
Ah dooh, I forgot. If you can add tests please do.

gaborbernat:
@ionelmc can we release this?

ionelmc:
Yes yes, I'll get it sorted this week.

blueyed:
@nedbat 
Maybe coverage.py should handle it similar to `COVERAGE_PROCESS_START` (i.e. pass it on, e.g. via `COVERAGE_CONTEXT`, which is used internally already however (metacov.ini, igor.py).

ionelmc:
Well I'm closer now to releasing, just waiting to sort out the travis-ci.com migration.

ionelmc:
`2.11.0` out now.

PR #450, opened 2021-01-05
Add in pre commit hook

EigenJT:
Added in `.pre-commit-hooks.yaml` to allow use with https://github.com/pre-commit/pre-commit

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+-   id: pytest-cov
+    name: pytest-cov
+    description: This plugin produces coverage reports. 
+    entry: pytest
+    language: python
+    types: [python]
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,3 +4,4 @@
     entry: pytest
     language: python
     types: [python]
+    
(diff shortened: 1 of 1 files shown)


ionelmc:
I guess this would only make it run a bare pytest? How would this work?

EigenJT:
Main use case I thought about was maintaining a required level of test
coverage before committing/pushing.

ionelmc:
Well ok. I'm not sure how this would be used. Some documentation is needed about the typical usage.

hukkinj1:
IMO this is not the right place for this hook. Should go to https://github.com/pytest-dev/pytest (if accepted by the maintainers) or a separate mirror repository.

The general pytest hook can then be extended with
```yaml
additional_dependencies: [pytest-cov]
args: [--cov, --cov-fail-under=95]
```
by the user, in order to add pytest-cov and coverage checking on top.

Even then this is probably a bad configuration because the user will need to list each and every project dependency in `additional_dependencies` to make the tests work...

graingert:
pre-commit hooks run in an isolated virtual environment, for pytest you need to be able to install your test extra deps for your tests to work

ssbarnea:
Sorry but pre-commit linting step has nothing to so with producing coverage reports. Running tests is a time consuming operation, clearly not something you want to bundle with coverage.

PR #459, opened 2021-03-25
Implement --cov-reset option that resets accumulated --cov directorie…

dsegan:
Implement --cov-reset option that resets accumulated --cov directories.

Don't seem to be able to link to an issue, so here it is: https://github.com/pytest-dev/pytest-cov/issues/460

FWIW, the tests here are the integration tests in line with the rest of them, though we could arguably have a simple (and much faster) unit test that just checks if the options are modified correctly.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -66,6 +66,8 @@ The complete list of command line options is:
                         False
   --no-cov              Disable coverage report completely (useful for
                         debuggers). Default: False
+  --cov-reset           Reset cov sources accumulated in options so far.
+                        Mostly useful for scripts and configuration files.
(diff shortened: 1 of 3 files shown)


dsegan:
Don't seem to be able to link this to an issue, so here it is: https://github.com/pytest-dev/pytest-cov/issues/460

dsegan:
Hi, thanks for taking a look at this earlier @ionelmc: and here's another friendly ping to get it merged if it looks fine :)

If there's something I need to do to get it merged, do let me know (eg. signing off on something).

ionelmc:
Nop, I just forgot about this.

dsegan:
Thanks!

Issue #464, opened 2021-04-25
multiprocess support


edcarter:
# Summary

Would you accept a patch to support the multiprocess library? https://pypi.org/project/multiprocess/

Currently the hooks in [embed.py](https://github.com/pytest-dev/pytest-cov/blob/master/src/pytest_cov/embed.py#L36) only install hooks for the built in multiprocessing library. I have written a small patch which adds the hooks for the multiprocess library as well so we can get code coverage for our pytests which use the multiprocess  library. The patch is relatively small and simple.

ionelmc:
Sure, open up a PR, I'm really curious now :-)

edcarter:
Alright, I will open up a PR tomorrow and let you see what code changes are needed.

Currently, I haven't updated any of the documentation or tests -- those will be the majority of the work needed to be done, but at least I can share the code first.

edcarter:
Better late than never. I created the pull request for you to take a look at. It is really quite simple

PR #466, opened 2021-05-03
WIP: Multiprocess support

edcarter:
Add support for multiprocess library as outlined in: https://github.com/pytest-dev/pytest-cov/issues/464

This is a work in progress since no documentation or tests have been updated.


ionelmc:
Seems simple enough. I guess for the tests ... some parametrization on the multiprocessing tests (parametrize which implementation) would be the way to do it?

ionelmc:
I think we can't do this anymore  (mp support is gone, users have to rely on whatever coveragepy has for that).

PR #467, opened 2021-05-12
change license in setup.py from BSD to MIT to match LICENSE file

csreesan:
License listed in setup.py says BSD while LICENSE file says MIT.
This led some tools  for license check to incorrectly state that the license for pytest-cov is BSD.
Changing setup.py license from BSD to MIT to correct this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -102,7 +102,7 @@ setup(
         'Development Status :: 5 - Production/Stable',
         'Framework :: Pytest',
         'Intended Audience :: Developers',
-        'License :: OSI Approved :: BSD License',
+        'License :: OSI Approved :: MIT License',
         'Operating System :: Microsoft :: Windows',
         'Operating System :: POSIX',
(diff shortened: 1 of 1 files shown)


sschuberth:
> Changing setup.py license from BSD to MIT to correct this.

For the record, could one of the maintainers confirm that no version of this library was never BSD licensed, and that this fix should also apply to any past releases of this library?

sschuberth:
> this library was never BSD licensed

[LICENSE file history](https://github.com/pytest-dev/pytest-cov/commits/master/LICENSE) shows that it has always been MIT since the [inception of the project](https://github.com/pytest-dev/pytest-cov/commit/5bacf2a65ce40f628f7802894fa309f24fccca04).

ionelmc:
Yeah looks like 2.11.1 and older releases  where published with incorrect metadata (classifiers had BSD). However, the license field had the correct value (MIT).

PR #470, opened 2021-05-13
fix link in readme

terencehonles:


ionelmc:
OOof, damn it - another thing I messed up with the bumpversion config. Thanks for finding this.

terencehonles:
No prob, I tried to use a replacement directive to make it one place to edit, but that didn't work so I figured I'd at least bump them to the same number so that a search would turn up both locations.

PR #481, opened 2021-07-30
Add Python 3.9 to trove classifiers

D3X:


ionelmc:
Well technically we still have 1 python 3.5 entry in the travis conf so it's tested.

D3X:
@ionelmc Right. Should I add it back or remove it from travis? 3.5's been EOL for almost a year now.

hugovk:
https://github.com/pytest-dev/pytest-cov/pull/488 is dropping 3.5, so I guess it doesn't matter?

PR #508, opened 2021-12-03
Remove forced config file default of .coveragerc

ofek:
There has been support for `pyproject.toml` ever since 5.0 https://github.com/nedbat/coveragepy/issues/664#issuecomment-552516277 and based on https://coverage.readthedocs.io/en/6.2/config.html it will automatically find the right file (logic here: https://github.com/nedbat/coveragepy/blob/6.2/coverage/config.py#L493-L519)

I tested this locally and now config can properly be read from the other fallback files


ofek:
@ionelmc @graingert Hello! When you get a moment can you please take a look at this? I can't figure out why the failures are occurring.

ionelmc:
Ooof sorry for delay, CI approved now. At first glance it seem fine. Please also update AUTHORS/CHANGELOG.rst

ofek:
Done! I don't know how to fix the tests though.

ofek:
Can I please get help with this?

ionelmc:
Hello, sorry for tardy response - can you rebase on the master? The CI was cleaned up.

ofek:
I'm not sure how to fix

ofek:
I still can't figure it out

src/pytest_cov/engine.py
```diff
@@ -98,7 +102,7 @@ def set_env(self):
         if os.path.exists(config_file):
             os.environ['COV_CORE_CONFIG'] = config_file
         else:
-            os.environ['COV_CORE_CONFIG'] = os.pathsep
+            os.environ['COV_CORE_CONFIG'] = ''
```
ionelmc, line 105: I believe this change is the cause of all the regressions.

Note that we have this in embed.py and you need to keep it working:
```
        if cov_config == os.pathsep:
            cov_config = True
```

src/pytest_cov/engine.py
```diff
@@ -70,6 +70,10 @@ def __init__(self, cov_source, cov_report, cov_config, cov_append, cov_branch, c
         self.topdir = os.getcwd()
         self.is_collocated = None
 
+        # If unset (the default), indicate fallback behavior for other files like pyproject.toml.
+        # See https://github.com/nedbat/coveragepy/blob/6.2/coverage/control.py#L144-L146
+        self.config_file_choice = self.cov_config or True
```
ionelmc, line 75: Also I'd simply rename this to config_file. Maybe it's just me but a dropdrown widget pops up in my head when I see "choice" - too much Django I guess xD

ofek:
Can you please help?

webknjaz:
@ofek this looks old and needs conflict resolution. Try rebasing once more?

ofek:
CI was failing and I don't think the maintainer has time for this unfortunately so there's no point. If you really want me to I will but I feel like it's a waste of time.

ionelmc:
Sorry, I know this feels bad, having to rebase over and over. But I did point in code commends at the potential cause of the regressions. In hindsight I should have asked you to describe more clearly what this PR is supposed to do and help figure out a good solution instead of reviewing a solution that I am not sure what is supposed to fix. Maybe that would had gotten things moving faster.

Can we restart this? Is there a problem with pytest-cov loading the coverage configuration from a pyproject.toml?

PR #511, opened 2021-12-11
`--cov-fail-under` should not cause `pytest --collect-only` to fail

### LINKED ISSUE #426: pytest --collect-only fails when 'fail under' is set on a configuration
# Summary
My pyproject.toml have the following sections (among other):
```toml
[tool.pytest.ini_options]
minversion = '6.0'
addopts = '--quiet --cov'

[tool.coverage.run]
source = ['my_module']

[tool.coverage.report]
fail_under = 75
```
 
And some tools (such as VSCode Python's plugin) depends on `pytest --collect-only`. However, when running that it fails as it doesn't actually run the tests and it displays a coverage way below the real value.
 
## Expected vs actual result
Expected:
when running `pytest --collect-only` I'd expect it only to fail if for some reason it collect the list of tests, ignoring the fail under parameter.
 
Actual result:
`pytest --collect-only` fails stating the coverage didn't reach the goal.
 
# Reproducer
```
git clone https://github.com/maurofaccenda/pytest-cov-issue.git
cd pytest-cov-issue/
python3 -m venv venv
source ./venv/bin/activate
pip install pip --upgrade
pip install -r requirements.txt
pytest  # succeeds
pytest --collect-only  # fails
echo $?  # prints 1
```
 
## Versions

`pip list` output:
```
Package        Version
-------------- -------
attrs          20.1.0
coverage       5.2.1
iniconfig      1.0.1
more-itertools 8.4.0
packaging      20.4
pip            20.2.2
pkg-resources  0.0.0
pluggy         0.13.1
py             1.9.0
pyparsing      2.4.7
pytest         6.0.1
pytest-cov     2.10.1
setuptools     44.0.0
six            1.15.0
toml           0.10.1
```

```
$ python --version
Python 3.8.2
```

```
$ pytest --version
pytest 6.0.1
```

## Config
```toml
[tool.pytest.ini_options]
minversion = '6.0'
addopts = '--quiet --cov'

[tool.coverage.run]
source = ['my_module']

[tool.coverage.report]
fail_under = 75
```

## Code

https://github.com/maurofaccenda/pytest-cov-issue

Zac-HD:
I often have `--cov-fail-under=100` set from a config file, because I expect full coverage from running my tests.

With [HypoFuzz](https://pypi.org/project/hypofuzz/), however, I use `pytest --collect-only` to collect the tests and then run them with the fuzzer... but `--cov-fail-under` returns an error because collection didn't reach 100% coverage!  That seems pretty silly, so this patch makes us skip the coverage check if pytest is running in `--collect-only` mode.  Fixes #426. 

(I can `hypothesis fuzz -- --cov-fail-under=0` for now, but that's a pretty inelegant workaround which I'd rather not require of my users)

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -50,3 +50,4 @@ Authors
 * Brian Rutledge - https://github.com/bhrutledge
 * Danilo Šegan - https://github.com/dsegan
 * Michał Bielawski - https://github.com/D3X
+* Zac Hatfield-Dodds - https://github.com/Zac-HD
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,14 @@ Changelog
 =========
 
 
+3.1.0 (future)
+-------------------
+
+* `--cov-fail-under` no longer causes `pytest --collect-only` to fail
+  Contributed by Zac Hatfield-Dodds in
+  `#511 <https://github.com/pytest-dev/pytest-cov/pull/511>`_.
+
+
 3.0.0 (2021-10-04)
 -------------------
 
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -308,7 +308,7 @@ class CovPlugin:
                 warnings.warn(CovReportWarning(message))
                 self.cov_total = 0
             assert self.cov_total is not None, 'Test coverage should never be `None`'
-            if self._failed_cov_total():
+            if self._failed_cov_total() and not self.options.collectonly:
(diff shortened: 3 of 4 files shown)


ionelmc (approved):
Looks good.

Zac-HD:
@ionelmc - is there anything I can do to help get this merged?

I've just had a couple of complaints from users which I eventually traced back to this, and then worked out that I'd already made a PR.  But it doesn't help anyone until it's released 😭

Issue #535, opened 2022-05-23
add lcov as output format


alexises:
I'm currently try to setup automatic code coverage into vscode. the only available plugin require that coverage output is set as lcov format.

This format support had been integrated recently into coverage lib :  https://coverage.readthedocs.io/en/6.3.2/cmd.html#lcov-reporting-coverage-lcov and I need it into pytest-cov for a good integration

This feature seen to be quite simple to implement : 

 * from `def validate_report(arg):` we should add the new format into `file_choices`
 * from ` def summary(self, stream):` on engine, we should add a block for lcov format like the one for xml on line 193-197

I'm not sure if it's fullfill all the needed work, but it's a good start

nedbat:
You don't need to add anything to pytest-cov.  Just run `coverage lcov` after you run pytest.

alexises:
Technically it's work, but with the integration with vscode it's require one manual step.

My main interrogation is why it could be an issue to add another output format. Let the code as this break expectation that all format supported inside coverage library is supported inside pytest-cov

PR #536, opened 2022-05-27
Add support for LCOV output

fetzerch:
Coverage.py 6.3 gained support for the LCOV output format. Add support for this to *pytest-cov* via `--cov-report=lcov[:dest]`.

Fix: #535

My main motivation was that coveralls currently only supports LCOV in their [GitHub Action](https://github.com/coverallsapp/github-action/issues/30). But since LCOV is a widely adopted format I think there might be more benefits of adding this to *pytest-cov*.

I've tried to follow [CONTRIBUTING.rst](https://github.com/pytest-dev/pytest-cov/blob/master/CONTRIBUTING.rst), let me know if you want me to add/change anything.

~*Update:* The second commit is unrelated, but fixes the check pipeline.~

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -51,3 +51,4 @@ Authors
 * Danilo Šegan - https://github.com/dsegan
 * Michał Bielawski - https://github.com/D3X
 * Zac Hatfield-Dodds - https://github.com/Zac-HD
+* Christian Fetzer - https://github.com/fetzerch
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,9 @@ Changelog
     concurrency = multiprocessing
     parallel = true
     sigterm = true
+* Added support for LCOV output format via `--cov-report=lcov`. Only works with coverage 6.3+.
+  Contributed by Christian Fetzer in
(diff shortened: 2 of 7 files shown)


ionelmc:
Hello, sorry for tardy response - can you rebase on the master? The CI was cleaned up.

ssbarnea:
@ionelmc if I remember correctly there were some discussions if this plugins should even do any kind of reporting instead of relying on coverage.py itself (to avoid feature duplication and extra complexity).

ionelmc:
There was but unless we remove all the reporting we can't really say "no we can't support another reporter" since it would be so inconsistent in both features and policy.

ssbarnea:
If we look to remove them, we should first stop adding new ones, especially when these can easily be covered by coverage tool itself. That would avoid having someone invest time trying to implement a feature that will (hopefully soon) be removed. Better to tell people to use a more future-proof way of achieving their goal.

MarximusMaximus:
Isn't this just telling coverage.py to generate the lcov.xml file for us? All I use pytest-cov for is to tell coverage.py to do things for me when running pytest. Is there another point to this plugin other than that?

ionelmc:
Yeah we should stop adding stuff but if we don't have reporting parity it makes pytest-cov less desirable. This doesn't look like it complicates things a lot anyway.

PR #549, opened 2022-07-31
Use modern approach to specify hook options

nicoddemus:
The old way using marks is being deprecated in pytest 7.2:

https://github.com/pytest-dev/pytest/pull/9118

### DIFF
--- a/src/pytest_cov/compat.py
+++ b/src/pytest_cov/compat.py
@@ -3,15 +3,8 @@ try:
 except ImportError:
     from io import StringIO
 
-import pytest
 
 StringIO  # pyflakes, this is for re-export
-
-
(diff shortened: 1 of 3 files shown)


src/pytest_cov/compat.py
```diff
@@ -3,17 +3,10 @@
 except ImportError:
     from io import StringIO
 
-import pytest
 
 StringIO  # pyflakes, this is for re-export
 
 
-if hasattr(pytest, 'hookimpl'):
-    hookwrapper = pytest.hookimpl(hookwrapper=True)
```
nicoddemus, line 12: `hookimpl` has been introduced in [pytest 2.8](https://docs.pytest.org/en/stable/changelog.html#id390), so this compatibility can be dropped.

PR #551, opened 2022-08-26
Replace `setuptools` build backend with `hatchling`

ofek:
cc @graingert as requested!

## Background

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://peps.python.org/pep-0517/)/[PEP 660](https://peps.python.org/pep-0660/)) and the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)). As a result, the execution of `setup.py` files is now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

## Summary of changes

This implements [PEP 621](https://peps.python.org/pep-0621/), obviating the need for `setup.py` and `MANIFEST.in`. The build backend `hatchling` (of which I am a maintainer in the [PyPA](https://github.com/pypa)) is now used as that is the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Hatchling is available on all the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hatchling/default.nix), [Alpine Linux](https://github.com/alpinelinux/aports/blob/master/community/py3-hatchling/APKBUILD), [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-hatchling/Makefile)/[OpenBSD](https://github.com/openbsd/ports/blob/master/devel/py-hatchling/Makefile), [Gentoo Linux](https://github.com/gentoo/gentoo/tree/master/dev-python/hatchling), [MacPorts](https://github.com/macports/macports-ports/blob/master/python/py-hatchling/Portfile), [OpenEmbedded](https://github.com/openembedded/openembedded-core/commit/846e806181f1349be29cbce78c5041735dfd7e6f), [Spack](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-hatchling/package.py), [MSYS2](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-hatchling/PKGBUILD), etc.

### Python support

The earliest supported Python 3 version of Hatchling is 3.7, therefore I've also set that as the minimum here. Python 3.6 is EOL and other build backends like `flit-core` and `setuptools` also dropped support. Python 3.6 accounted for 6-8% of downloads in the last month: https://pypistats.org/packages/pytest-cov

## Notes

- The source distributions on PyPI are erroneously shipping a build artifact `*.egg-info` from `python setup.py develop`; this is now fixed
- The `CHANGELOG.rst` regex substitution was doing nothing so the PyPI landing page continues to be simply `README.rst` + `CHANGELOG.rst`
- The generated `pytest-cov.pth` was being checked into the repo but is not necessary now so I've removed it

## Future

- Move config for tools like `pytest` out of `setup.cfg` and into `pyproject.toml`


ionelmc:
What exactly is `hatch-autorun`? How does it work? I don't get how this replaces the pth file we use.

ofek:
It essentially does the current logic of executing the `src/pytest-cov.embed` in a PTH file.

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+  "Topic :: Software Development :: Testing",
+  "Topic :: Utilities",
+]
+dependencies = [
+  "coverage[toml]>=5.2.1",
+  "pytest>=4.6",
+]
+
+[project.optional-dependencies]
+testing = [
```
ionelmc, line 49: This works as the "testing" extra right?

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+[project.entry-points.pytest11]
+pytest_cov = "pytest_cov.plugin"
+
+[project.urls]
+Changelog = "https://pytest-cov.readthedocs.io/en/latest/changelog.html"
+Documentation = "https://pytest-cov.readthedocs.io/"
+Homepage = "https://github.com/pytest-dev/pytest-cov"
+"Issue Tracker" = "https://github.com/pytest-dev/pytest-cov/issues"
+
+[tool.hatch.metadata.hooks.fancy-pypi-readme]
```
ionelmc, line 67: Does the fancy pypi readme do the ref tag stripping automatically? The ol setup.py had a ```re.sub(':[a-z]+:`~?(.*?)`', r'``\1``', ...```

pyproject.toml
ofek: Yup!

pyproject.toml
ofek: No, but as I mentioned in the OP it was doing nothing:

```pycon
Python 3.11.0b3 (main, Jun  1 2022, 13:29:14) [MSC v.1932 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pathlib, re
>>>
>>> before = pathlib.Path('CHANGELOG.rst').read_text(encoding='utf-8')
>>> after = re.sub(':[a-z]+:`~?(.*?)`', r'``\1``', before)
>>>
>>> before == after
True
```

ofek:
CI passing btw

ofek:
Just let me know if I can help!

ionelmc:
I'll be honest here I'm not a fan of redoing something that works. But I am also curious how well this could work so how about this: release it as pytest-cov 5.0.0a (or beta) and we keep this in a special "alpha" branch till people try it out enough? This can only work if you pledge yourself to maintain pytest-cov's packaging from here on out :-)

RonnyPfannschmidt:
I'm also in on this, starting with pluggy, then going for other pytest projects

I like hatch enough that i decided to let my own tool die in favour of it.

ofek:
I rebased to fix conflicts. Would you mind triggering the CI with that button below?

ofek:
I rebased again, accounting for https://github.com/pytest-dev/pytest-cov/pull/553#issuecomment-1284912148

ofek:
flake?

ofek:
I rebased again, accounting for https://github.com/pytest-dev/pytest-cov/pull/556

ofek:
Would you mind triggering the CI with that button below?

ofek:
Passing 🙂

cclauss:
Please rebase.

ofek:
I don't know whether to invest time rebasing because of the previous lack of correspondence.

docs/releasing.rst
```diff
@@ -21,12 +21,12 @@ The process for releasing should follow these steps:
 #. Make sure you have a clean checkout, run ``git status`` to verify.
 #. Manually clean temporary files (that are ignored and won't show up in ``git status``)::
 
-        rm -rf dist build src/*.egg-info
+        rm -rf dist
```
webknjaz, line 24: Perhaps, this could be replaced by a `git clean` which is cross-platform.

ionelmc:
Looks like there's no action on this so I took the liberty to rebase it in #716.

PR #553, opened 2022-09-16
Enable `pytest-cov[toml]` akin to upstream `coverage[toml]`

stdedos:
https://github.com/nedbat/coveragepy/issues/664 (+ later fixes)

Signed-off-by: Stavros Ntentos <133706+stdedos@users.noreply.github.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -142,7 +142,9 @@ setup(
             'six',
             'pytest-xdist',
             'virtualenv',
-        ]
+        ],
+        # Enable pyproject.toml support.
+        'toml': ['coverage[toml]'],
     },
     entry_points={
         'pytest11': [


setup.py
```diff
@@ -142,7 +142,9 @@ def run(self):
             'six',
             'pytest-xdist',
             'virtualenv',
-        ]
+        ],
+        # Enable pyproject.toml support.
+        'toml': ['tomli; python_full_version<="3.11.0a6"'],
```
ionelmc, line 147: It seems to me like this would have been written like `'toml': ['coverage[toml]'],` instead (so I don't need to worry about updating this ever).

stdedos: Seems to me you are right :-D

Thanks for the suggestion!

ofek:
Related https://github.com/pytest-dev/pytest-cov/pull/508

stdedos:
@ionelmc all good now? :confused:

ofek:
Wait, what's the point of this? We already do that:

https://github.com/pytest-dev/pytest-cov/blob/f7bdfcfe3aae34982a955b8609cd345820f2526e/setup.py#L132-L135

stdedos:
When I tried to move .coveragerc to pyproject.toml, I was unable to read the configuration. 

I assumed that this was a necessary step (and I was going to test if this fixed the issue after a new pypi release).

ofek:
Without https://github.com/pytest-dev/pytest-cov/pull/508 you need `--cov-config=pyproject.toml`

stdedos:
> Without https://github.com/pytest-dev/pytest-cov/pull/508 you need `--cov-config=pyproject.toml`

Oh well 🤦

Good to know; I'll try to check this on my new project sometime.

ionelmc:
Facepalm.

PR #556, opened 2022-10-20
Revert "Enable `pytest-cov[toml]` akin to upstream `coverage[toml]`"

ionelmc:
Reverts pytest-dev/pytest-cov#553


ofek:
fyi I also removed it in https://github.com/pytest-dev/pytest-cov/pull/551#issuecomment-1284914598

Issue #557, opened 2022-10-26
pytest-cov is incompatible with pytest-xdist 3.x due to injection of rsync option


ssbarnea:
# Summary

See https://github.com/pytest-dev/pytest-xdist/issues/825#issuecomment-1292306864 which reports a deprecation option related to rsyncdir(s) options. This breaks the builds for anyone that run with warnings-as-errors.

## Code

https://github.com/pytest-dev/pytest-cov/blob/f7fced579e36b72b57e14768026467e4c4511a40/src/pytest_cov/engine.py#L263
Link to your repository, gist, pastebin or just paste raw code that illustrates the issue.

## Solution

We need to stop adding this option on newer versions as it was already deprecated and will be removed on next version of xdist.

DanielNoord:
We're still seeing this `DeprecationWarning` being raised on the new release. What can still trigger it after the PR got merged?

psi29a:
We're also still getting this message.

Our workaround was to not use --cov-config and instead pass the env variable `COVERAGE_RCFILE`.

```
COVERAGE_RCFILE=.coveragerc pytest --cov
```

ionelmc:
Oh damn I worded the changelog incorrectly. There's a feature test in pytest-cov for rsyncdir support. Basically the change allows pytest-cov to continute to work with xdist 4, it doesn't deal with the warning.

psi29a:
It's strange, because passing this env variable tells coverage to use the file I want... if we use this:
```
pytest -n logical --maxprocesses=4 --dist loadgroup --cov --cov-config=../configuration/.coveragerc --cov-append --cov-branch
```

We get this:
```
INTERNALERROR> DeprecationWarning: The --rsyncdir command line argument and rsyncdirs config variable are deprecated.

INTERNALERROR> The rsync feature will be removed in pytest-xdist 4.0.
```

This does not make sense, as we don't use rsync at all.

Doing this however:
```
COVERAGE_RCFILE=.coveragerc pytest -n logical --maxprocesses=4 --dist loadgroup --cov --cov-append --cov-branch
```
then we can run without issues and no warning.

Seems like an issue to me, can you verify please? :)

ionelmc:
Works as expected - because you have specified the config file via pytest-cov and enabled xdist it's being sent over xdist's file sharing mechanism. It may not be what you want but it is what it's implemented.

mdantonio:
> it doesn't deal with the warning

I'm a bit confused here 😖
[...]

PR #558, opened 2022-10-26
Remove use of rsyncdir

ssbarnea:
Fixes: #557

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -260,7 +260,9 @@ class DistMaster(CovController):
 
         # Ensure coverage rc file rsynced if appropriate.
         if self.cov_config and os.path.exists(self.cov_config):
-            self.config.option.rsyncdir.append(self.cov_config)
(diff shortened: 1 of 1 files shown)


RonnyPfannschmidt (changes requested):
That particular example and the loops are actually correct and needed as is

We might need to revise the rsync dir details given thst we don't have a coverage combine test atm

ionelmc:
I'd remove it completely now. Best to deal with the breakage now than down the road when support for that option is gone.

ssbarnea:
@ionelmc Any chance to approve it?

tomhamiltonstubber:
Any word here folks?

ionelmc:
Whoops missed this. Looks all good.

jonyscathe:
Any chance of a pytest-cov to get this fix in?
Would be great to git rid of deprecation warnings filters in all of my projects and this was fixed up months ago in master.

tomhamiltonstubber:
Please can we get a release here?

ionelmc:
Later today there will be a 4.1 release (if no objections).

PR #565, opened 2022-11-15
disabling boxed test if version xdist newer than 2.5.0

Chedi:
Newer version of xdist  (> 2.5.0) have removed the boxed option.

[xdist changelog](https://github.com/pytest-dev/pytest-xdist/blob/master/CHANGELOG.rst#pytest-xdist-250-2021-12-10)

### DIFF
--- a/tests/test_pytest_cov.py
+++ b/tests/test_pytest_cov.py
@@ -1555,7 +1555,8 @@ def test_foo():
 
 SCRIPT_SIMPLE_RESULT = '4 * 100%'
 
-
+@pytest.mark.skipif('tuple(map(int, xdist.__version__.split("."))) >= (2, 5, 0)',
+                    reason="--boxed option was removed in version 2.5.0")
 @pytest.mark.skipif('sys.platform == "win32"')
 def test_dist_boxed(testdir):
     script = testdir.makepyfile(SCRIPT_SIMPLE)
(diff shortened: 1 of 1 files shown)


cclauss (approved):

tests/test_pytest_cov.py
```diff
@@ -1555,7 +1555,8 @@ def test_foo():
 
 SCRIPT_SIMPLE_RESULT = '4 * 100%'
 
-
+@pytest.mark.skipif('tuple(map(int, xdist.__version__.split("."))) >= (2, 5, 0)',
+                    reason="--boxed option was removed in version 2.5.0")
```
hugovk, line 1559: It was deprecated in [2.5.0](https://github.com/pytest-dev/pytest-xdist/blob/master/CHANGELOG.rst#pytest-xdist-250-2021-12-10) and removed in [3.0.2](https://github.com/pytest-dev/pytest-xdist/blob/master/CHANGELOG.rst#pytest-xdist-302-2022-10-25).

```suggestion
@pytest.mark.skipif('tuple(map(int, xdist.__version__.split("."))) >= (3, 0, 2)',
                    reason="--boxed option was removed in version 3.0.2")
```

hugovk (changes requested):

ionelmc:
Rebased in master. Will release in 4.1 later today if no objections.

PR #592, opened 2023-04-30
Update README.rst

Jylpah:
I propose adding clarification of the abbreviations in the coverage report. These are not mentioned anywhere in the documentation and there is even a [StackOverflow question](https://stackoverflow.com/questions/55293195/what-is-pytest-result-mean )about those. 

I can understand these are "obvious" for the experienced programmers, but maybe not for someone taking a code coverage tool into use for the first time. 

I understand you want to keep the documentation short and to the point (it is).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,6 +117,11 @@ Would produce a report like::
     ----------------------------------------
     TOTAL                  353     20    94%
 
+Where, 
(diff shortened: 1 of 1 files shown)


nedbat:
Another option: add information like this to the coverage.py docs, with a clear indication from here that much of the pytest-cov behavior is from coverage.py.

PR #595, opened 2023-05-17
update documentation copyright year

Samoed:


RonnyPfannschmidt (changes requested):
Having that value dynamic is fundamentally wrong, it is either a start year, or a definite range, never the current date

PR #611, opened 2023-09-19
Removing message formatting rounding

didibz-harmonya:
When defining a coverage threshold (88.45 for example) and the actual coverage is very close (88.449  for example) we will get a failure message. However, the message will be: "FAIL Required test coverage of 88.45% not reached. Total coverage: 88.45%" where it's wrong and confusing since we failed because it's under while it the message it presented as equal. The removing of the rounding will fix it.

### DIFF
--- a/src/pytest_cov/plugin.py
+++ b/src/pytest_cov/plugin.py
@@ -341,7 +341,7 @@ class CovPlugin:
             markup = {'red': True, 'bold': True} if failed else {'green': True}
             message = (
                 '{fail}Required test coverage of {required}% {reached}. '
-                'Total coverage: {actual:.2f}%\n'
+                'Total coverage: {actual}%\n'
                 .format(
                     required=self.options.cov_fail_under,
                     actual=self.cov_total,


AdiKrasin (approved):

Colin-b:
Hello @AdiKrasin, can this fix get merged and released?

Colin-b:
@ionelmc Can you have a look at this? Thanks again

Colin-b:
If you are not ok with this change, I'll gladly come up with a PR fixing it as well matching your expectations. This is a long standing issue that is quite painful with long running test suites (because it's a pain to relaunch it to know the coverage) that are distributed (so cannot run with coverage only). 

I'd argue a more in-line change would be to read the coverage "precision" setting and report it the same way coverage is. Is that something you would be willing to merge and release?

src/pytest_cov/plugin.py
```diff
@@ -341,7 +341,7 @@ def pytest_terminal_summary(self, terminalreporter):
             markup = {'red': True, 'bold': True} if failed else {'green': True}
             message = (
                 '{fail}Required test coverage of {required}% {reached}. '
-                'Total coverage: {actual:.2f}%\n'
+                'Total coverage: {actual}%\n'
```
Pierre-Sassoulas, line 344: I think this would fix the problem you described without changing the output too dramatically.
```suggestion
                'Total coverage: {truncate(actual, 2)}%\n'
```

Colin-b: Would be cleaner imo to rely on what coverage is doing: https://github.com/nedbat/coveragepy/blob/a1d8d2961c9793cf8d3fa09666012190cdec9214/coverage/results.py#L253

Pierre-Sassoulas: Right !

ionelmc: Maybe we could just reuse that from coverage? Hopefully @nedbat won't change it soon :-)

nedbat: I am actively refactoring that file to make more of it public: https://github.com/nedbat/coveragepy/tree/nedbat/public-analysis . We could talk about how to make it useful here.

Issue #621, opened 2023-11-28
Remove usage of deprecated `rsyncdir`, which will be removed in pytest-xdist 4.0


parthea:
# Summary

This issue is related to https://github.com/pytest-dev/pytest-cov/issues/557 which is closed.

The `rsyncdir` argument is deprecated and will be removed in pytest-xdist 4.0. It's currently used in the line below:
https://github.com/pytest-dev/pytest-cov/blob/2c9f2170d8575b21bafb6402eb30ca7de31e20b9/src/pytest_cov/engine.py#L272

See other usage here:
https://github.com/search?q=repo%3Apytest-dev%2Fpytest-cov%20rsyncdir&type=code

## Expected vs actual result
When using `pytest-cov` I didn't expect to see warnings. A warning appears due to usage of `rsyncdir` which is deprecated.

```
py.test -n=auto --cov=test
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/local/google/home/partheniou/.pyenv/versions/3.12.0/envs/testpycov/lib/python3.12/site-packages/_pytest/main.py", line 267, in wrap_session
INTERNALERROR>     config._do_configure()
INTERNALERROR>   File "/usr/local/google/home/partheniou/.pyenv/versions/3.12.0/envs/testpycov/lib/python3.12/site-packages/_pytest/config/__init__.py", line 1053, in _do_configure
INTERNALERROR>     self.hook.pytest_configure.call_historic(kwargs=dict(config=self))
INTERNALERROR>   File "/usr/local/google/home/partheniou/.pyenv/versions/3.12.0/envs/testpycov/lib/python3.12/site-packages/pluggy/_hooks.py", line 514, in call_historic
INTERNALERROR>     res = self._hookexec(self.name, self._hookimpls, kwargs, False)
INTERNALERROR>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/usr/local/google/home/partheniou/.pyenv/versions/3.12.0/envs/testpycov/lib/python3.12/site-packages/pluggy/_manager.py", line 115, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook_name, methods, kwargs, firstresult)
[...]

PR #623, opened 2023-11-28
remove rsyncdir usage since pytest-xdist has deprecated it

matthias-margin:
This removes the usage of rsyncdir so that pytest-xdist does not raise a deprecation warning. Trying to get all warning in my projects to go away. 

Issue submitted a bit ago https://github.com/pytest-dev/pytest-cov/issues/621 and a older one too https://github.com/pytest-dev/pytest-cov/issues/557

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -265,12 +265,6 @@ class DistMaster(CovController):
     def start(self):
         cleanup()
 
-        # Ensure coverage rc file rsynced if appropriate.
-        if self.cov_config and os.path.exists(self.cov_config):
-            # rsyncdir is going away in pytest-xdist 4.0, already deprecated
-            if hasattr(self.config.option, 'rsyncdir'):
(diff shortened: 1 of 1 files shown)


lesteve:
@ionelmc it seems like this should be merged. You said yourself in https://github.com/pytest-dev/pytest-cov/issues/557#issuecomment-1579332066

> Well I guess there could be another pytest-cov release outright removing the rsync stuff. Would that make everyone happy? 

And the response to your question seems to be a clear yes:
- ~10 thumbs up in response to your comment
- other comments going in the same direction in the same issue
- this PR that was opened to fix it and #621 that was opened as #557 to say the problem is still there

mscheltienne:
@ionelmc pinging as well, as a merge of this PR and a new release would be appreciated.

rschwebel:
Thanks for taking care about this! This patch is now in 5.0.0, but it hasn't been rolled out to pypi. Could someone take care of that?

ionelmc:
Yeah, this weekend I'll have time for it.

PR #629, opened 2024-02-15
Update tox example to properly track coverage

jongbinjung:
If `usedevelop=True` is not set, coverage for `tox` tests show up as "0%" because `tox` runs the tests in a virtualenv. (See https://stackoverflow.com/questions/58696476/tox-0-coverage)

### DIFF
--- a/examples/src-layout/tox.ini
+++ b/examples/src-layout/tox.ini
@@ -21,6 +21,7 @@ deps =
 
 depends =
     report: pypy3,py39
+usedevelop=True
 
 [testenv:report]
 skip_install = true


ionelmc:
I have ran the src example locally and it still produces the expected 100% coverage - is there something I am missing here?

webknjaz:
The Stackoverflow example seems to be missing the path mapping in the coveragepy config. This example shouldn't be updated. The patch is misleading.

jongbinjung:
Ah, I see. Got it. Thanks for clarifying.

PR #633, opened 2024-03-17
Fix or remove URLs that are causing docs tests to fail

cclauss:
Fix or remove URLs that are causing GitHub Actions `docs` tests to fail.

Tests docs passing

@schlamar @msabramo @joshk Your reviews, please.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -2,15 +2,15 @@
 Authors
 =======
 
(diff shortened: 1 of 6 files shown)


cclauss:
@ionelmc Your review, please.

ionelmc:
Partially applied in master (for the authors I just disabled the linking instead).

PR #640, opened 2024-04-04
Add .git-blame-ignore-revs

akx:
This makes GitHub (and other tools that support this file, or any Git client with `git config blame.ignoreRevsFile .git-blame-ignore-revs` set) ignore the large reformatting/automatic commits, to make blame inspection easier.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# introduced f-strings
+76fb2a6cb2c4a4a788a5b62710848daf9c8fb7ce
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -15,6 +15,7 @@ include .bumpversion.cfg
 include .cookiecutterrc
 include .coveragerc
 include .editorconfig
+include .git-blame-ignore-revs
 include .pre-commit-config.yaml
 include .readthedocs.yml
 include pytest.ini


Pierre-Sassoulas (approved):

akx:
@Pierre-Sassoulas Added the file to `MANIFEST.in` too to make the `check` Tox step pass.

Pierre-Sassoulas (changes requested):
76fb2a6cb2c4a4a788a5b62710848daf9c8fb7ce could be removed from the git blame but not ff3fdd8039bb11e93d7a5b06a32091d58e97bfbf which introduce some non formatting changes (remove six, maybe others) and some unsafe fixes from ruff. I'm pretty sure the migration from ``os.path`` to ``pathlib`` is not automated either (I initially thought it was because of this MR, until I actually tried and the deception was strong).

webknjaz:
I'd argue it's not just a formatting change but a functional one. F-strings are more performant so it's more of an enhancement rather than formatting. Usually only patches that are solely formatting are ignored. Bits that are useful for git paleontology aren't supposed to be hidden...

Pierre-Sassoulas (commented):
I'm not sure what's the point of `.git-blame-ignore-revs` in Manifest.in : after packaging git is not involved anymore, right ? So if f-strings removed from git blame are not consensual, there would be nothing left of this MR if we also remove that 😅

akx:
> I'm not sure what's the point of `.git-blame-ignore-revs` in Manifest.in

Without it there, `check-manifest` would fail.

Pierre-Sassoulas:
What is check-manifest and why does it fail ? manifest.in is linked to python packaging and does not involve git, right ?

ionelmc:
check-manifest is used to make sure the sdist contains all the source files and nothing is missing. Maybe this new file needs to be an exclude in there.

But alas, I am not sure who is this file for. I will not use it - I prefer to see everything in the gitblame, and jumping back a commit is pretty easy enough (it's just one button in github anyway).
image

webknjaz:
> check-manifest is used to make sure the sdist contains all the source files and nothing is missing. Maybe this new file needs to be an exclude in there.
> 
> But alas, I am not sure who is this file for. I will not use it - I prefer to see everything in the gitblame, and jumping back a commit is pretty easy enough (it's just one button in github anyway). image

@ionelmc this file is picked up by GitHub automatically, and it hides the listed commits from blame, unlike local Git invocations where that needs to be configured.

webknjaz:
> check-manifest is used to make sure the sdist contains all the source files and nothing is missing. Maybe this new file needs to be an exclude in there.

Have you considered using `setuptools-scm` which auto-includes everything Git-tracked into sdists?

webknjaz:
@ionelmc would you be open to integrating `setuptools-scm`?

ionelmc:
@webknjaz I am not opposed in principle. I use it on projects with C extensions where it's useful to have uniquely versioned pre-release distributions. But on this project I don't see the value to change an already working packaging setup. As a matter of fact I only see myself responding to bug reports caused by such a change.

Anyway, regarding this ignore-revs change - I don't want it. I simply don't trust even myself to not introduce behavior changes in formatting commits.

PR #645, opened 2024-05-27
Use ruff for formatting too

akx:
Following up on `ruff` being added for linting in #610, this makes formatting also use `ruff` instead of Black.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,14 +6,11 @@ exclude: '^(\.tox|ci/templates|\.bumpversion\.cfg)(/|$)'
 # Note the order is intentional to avoid multiple passes of the hooks
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.3.3
+    rev: v0.4.5
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix, --show-fixes, --unsafe-fixes]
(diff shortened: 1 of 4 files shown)


ionelmc (commented):
This might be a good time to replace `[tool.ruff.lint.flake8-quotes] inline-quotes = "single"` with `[tool.ruff.format] quote-style = "single"`.

This ruff formatter seems new, or I have been time traveling :)) - what is your experience with it?

Pierre-Sassoulas:
It's available since November 2023 afair. In m'y experience It's faster than black, and it also permits to not argue about formatting. (Equivalent at first but they started to drift apart after black's 2024 style.. two months after first release). As long as it's pre-commit enforced and you don't have strong opinion about following black's style, it should not be a problem and a small quality of life increase when committing or reading the pre-commit conf.

akx:
> This ruff formatter seems new, or I have been time traveling :)) - what is your experience with it?

It's pretty new, yeah, but it's been rock stable and fast (as all of Astral's tools seem to be...). 99% compatible with the Black style, with some changes that are IMO for the better.

I can switch the quote styles around, sure, but that'll probably be a much larger diff. So – enforce single quotes everywhere is the preference?

ionelmc:
Enforcing single quotes everywhere is what `[tool.ruff.lint.flake8-quotes] inline-quotes = "single"` was supposed to do. I guess I have to try this and see how it goes.

PR #647, opened 2024-06-03
Move testing extras to a requirements_test.txt file

mbyrnepr2:
Move the ``testing`` extras from the setup.py to their own requirements_test.txt file.

Please let me know if I'm missing some details; I've put this together rather quickly.

The rationale here is that the `extras` list was showing up in a poetry.lock file for a project at my company. This was because pytest-cov was part of the dev-dependencies in a pyproject.toml file.
Our vulnerability scanning tool is designed to skip development dependencies. However, it doesn't skip a development dependency if it appears in the extras list in the poetry.lock file. Edit: the previous line appears to be inaccurate - it is skipped if there are only dev dependencies in pyproject.toml. 

This is not a problem with pytest-cov per-se but perhaps the change is acceptable in any case and helps others potentially. Maybe this PR makes sense from the perspective that the test extras would never be used by downstream users, but otherwise please feel free to close it😅

Thanks for your time!

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -25,5 +25,6 @@ include CHANGELOG.rst
 include CONTRIBUTING.rst
 include LICENSE
 include README.rst
+include requirements_test.txt
 
 global-exclude *.py[cod] __pycache__/* *.so *.dylib
--- a/requirements_test.txt
+++ b/requirements_test.txt
@@ -1,0 +1,5 @@
+fields
+hunter
+process-tests
+pytest-xdist
+virtualenv
--- a/setup.py
+++ b/setup.py
@@ -129,15 +129,6 @@ setup(
         'pytest>=4.6',
         'coverage[toml]>=5.2.1',
     ],
-    extras_require={
-        'testing': [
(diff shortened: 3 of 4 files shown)


ionelmc:
@blueyed added those extras, not sure if he needs them. 

Regardless, I suspect this change will trip up linux packages for pytest-cov that use those extras for testing.

Why does poetry use that extra?

mbyrnepr2:
> @blueyed added those extras, not sure if he needs them.
> 
> Regardless, I suspect this change will trip up linux packages for pytest-cov that use those extras for testing.
> 
> Why does poetry use that extra?

Fair point - and I'm not convinced my approach improves the situation.
It looks like uv.lock and pdm.lock do not contain that extras data that poetry.lock does; but I guess that's a separate matter.

Thanks for looking into it @ionelmc; I appreciate it but there's probably nothing needed here; the pattern looks common enough anyway: https://discuss.python.org/t/is-it-preferable-to-add-test-dependensies-as-an-extra-require-instead-of-tox-ini/4984

PR #684, opened 2025-03-31
Make sure the CLI precision is used when creating report. Fixes #674.

### LINKED ISSUE #674: `--cov-precision` does not affect report
When you configure the following in `pyproject.toml`:

```toml
[tool.coverage.report]
precision = 2
```

you get the output:

```
Name                Stmts   Miss   Cover
----------------------------------------
mypkg/__init__.py       0      0 100.00%
mypkg/bar.py            4      4   0.00%
mypkg/foo.py            2      2   0.00%
----------------------------------------
TOTAL                   6      6   0.00%
```

However when you don't configure that and only use the recently introduced `--cov-precision` option you still get:

```
Name                Stmts   Miss  Cover
---------------------------------------
mypkg/__init__.py       0      0   100%
mypkg/bar.py            4      4     0%
mypkg/foo.py            2      2     0%
---------------------------------------
TOTAL                   6      6     0%
```

That is the `--cov-precision` option doesn't affect the generated reports at all.

ionelmc:


src/pytest_cov/engine.py
```diff
@@ -67,13 +67,16 @@ def _data_suffix(name):
 class CovController:
     """Base class for different plugin implementations."""
 
-    def __init__(self, cov_source, cov_report, cov_config, cov_append, cov_branch, config=None, nodeid=None):
+    cov: coverage.Coverage | None
```
webknjaz, line 70: Perhaps move this into the initializer? Any reason to have it right here?

src/pytest_cov/engine.py
```diff
@@ -67,13 +67,16 @@ def _data_suffix(name):
...
+
+    def __init__(self, cov_source, cov_report, cov_config, cov_append, cov_branch, cov_precision, config=None, nodeid=None):
```
webknjaz, line 72: It's probably best to make the new arg last + maybe kw-only so that it isn't a breaking change in the signature.

ionelmc: Well now that you mention it, it does look odd doesn't it. Such a large number of positional always leads to some trouble. Funny thing, the config and nodeid are always passed. It was like that since the first commit where this was set to be a common lib for both pytest and nose: https://github.com/pytest-dev/pytest-cov/commit/7c6448ddf699c20323068ee7c2a17aa60fd3a3ae

src/pytest_cov/engine.py
ionelmc: It's just a type annotation - ":" doesn't assign anything to the "cov" class attribute. I added it mostly to make some things easier in my editor.

Issue #693, opened 2025-06-02
coverage collection fails when using pytest-xdist and pytest 8.4.0


rjdbcm:
# Summary

pytest-xdist and pytest-cov plugins cause an internal error when used together but only on pytest version 8.4.0

## Expected vs actual result

Expected pytest-xdist+pytest-cov to succeed.
See the following chart for the actual results:

| ⬇ pytest ♦ plugins ➡ | pytest-xdist+pytest-cov | pytest-cov | pytest-xdist |
|-----------------|-------------------------|------------|--------------|
| pytest < 8.4.0  | ✅                      | ✅         | ✅           |
| pytest == 8.4.0 | :x:                     | ✅         | ✅           |

<details> 
<summary>
with the following output:
</summary>

```
...
created: 16/16 workers
16 workers [1 item]

scheduling tests via LoadScheduling

tests/test_metadata.py::test_metadata 
INTERNALERROR> def worker_internal_error(
INTERNALERROR>         self, node: WorkerController, formatted_error: str
INTERNALERROR>     ) -> None:
INTERNALERROR>         """
INTERNALERROR>         pytest_internalerror() was called on the worker.
INTERNALERROR>     
INTERNALERROR>         pytest_internalerror() arguments are an excinfo and an excrepr, which can't
INTERNALERROR>         be serialized, so we go with a poor man's solution of raising an exception
INTERNALERROR>         here ourselves using the formatted message.
INTERNALERROR>         """
INTERNALERROR>         self._active_nodes.remove(node)
INTERNALERROR>         try:
INTERNALERROR> >           assert False, formatted_error
INTERNALERROR> E           AssertionError: Traceback (most recent call last):
INTERNALERROR> E               File " /installdir/pluggy/_callers.py", line 43, in run_old_style_hookwrapper
INTERNALERROR> E                 teardown.send(result)
INTERNALERROR> E                 ~~~~~~~~~~~~~^^^^^^^^
INTERNALERROR> E               File " /installdir/pytest_cov/plugin.py", line 324, in pytest_runtestloop
INTERNALERROR> E                 self.cov_controller.finish()
INTERNALERROR> E                 ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
INTERNALERROR> E               File " /installdir/pytest_cov/engine.py", line 57, in ensure_topdir_wrapper
INTERNALERROR> E                 return meth(self, *args, **kwargs)
[...]

PR #695, opened 2025-06-06
Prevent unimported warnings @ pytest-xdist workers

webknjaz:
Ref #693

This fixes https://github.com/pytest-dev/pytest-cov/issues/693#issuecomment-2940497942 for me.
Actually, locally, I can also drop these private attribute assignments from `DistMaster` and it works with just disabling a single warning on the worker.
However, when I tried doing this in the project (and not in site-packages of where I'm hitting the problem), it breaks `tests/test_pytest_cov.py::test_xdist_no_data_collected` so I didn't include those changes.

I was hoping that you @ionelmc could take a look and maybe come up with some explanation of why this is happening.

Coverage is being collected and reported, but at the time `coveragepy` is looking into `sys.modules`, the module isn't there. I think xdist makes it unpredictable whether the coverage object in master or in the worker would hit this code path. It probably makes sense to disable this warning when under xdist for workers too. It's unreliable at best in such a setup.

### DIFF
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -450,6 +450,7 @@ class DistWorker(CovController):
             data_suffix=_data_suffix(f'w{self.nodeid}'),
             config_file=self.cov_config,
         )
+        self.cov._warn_unimported_source = False
         self.cov.start()
         self.set_env()
         super().start()


src/pytest_cov/engine.py
```diff
@@ -450,6 +450,7 @@ def start(self):
             data_suffix=_data_suffix(f'w{self.nodeid}'),
             config_file=self.cov_config,
         )
+        self.cov._warn_unimported_source = False
```
webknjaz, line 453: This should probably include all those other hacks, after all:
```suggestion
        self.cov._warn_no_data = False
        self.cov._warn_unimported_source = False
        self.cov._warn_preimported_source = False
```

webknjaz: Yeah, I only know that only one has a visible effect of suppressing a false negative in ansible-pylibssh. I've been trying to think of a reliable test but no luck so far.

PR #697, opened 2025-06-08
Fix document theme parameter in sphinx config file

koyuki7w:
Fixed the `html_theme_options` in config file since the option `githuburl` is not supported.
(ref: https://pradyunsg.me/furo/customisation/top-of-page-buttons/#with-popular-vcs-hosts .)
Now the "view/edit" buttons (on the top right) correctly point to the github repository.


ionelmc:
Thanks.

PR #706, opened 2025-07-20
Remove some extra indents in documentation

koyuki7w:
Incorrect indents cause some texts to be misinterpreted as blockquotes and render incorrectly.


ionelmc:
Thanks.

Issue #688, opened 2025-04-17
Add markdown output support


beck3905:
# Summary

I understand that pytest-cov uses coverage.py under the hood and coverage.py supports `--format=markdown` as an output option. Could this be added as an output format in pytest-cov?

This would be helpful adding the test report to a GitHub Actions Job Summary, which requires markdown format.

nedbat:
You can run `coverage report --format=markdown` as a second command and get the output without changing pytest-cov at all.

alelom:
We definitely could use this feature -- getting a markdown output would make it fast to display the coverage report from github, as a fast result from a github action.

webknjaz:
I think there's no harm in having this setting. Though, it's most useful within GHA when appending to the output file, not rewriting it. So `--cov-report markdown:"${GITHUB_STEP_SUMMARY}"` doesn't feel enough — it's unclear how one could specify that the file in the env var should be open in append mode?

nedbat:
@marcosboger I'm not a maintainer here so it's not my decision. And as I mentioned in my comment, I'm generally not in favor of jamming more features into this tool when the same thing can already be done in a simple way without adding maintenance burden here.

alelom:
> the same thing can already be done in a simple way without adding maintenance burden here.

I have yet to find a simple way to do this in my CI/CD workflows. If this feature was added in the same package it would be a game changer for me. Would you be able to share how can it be done in a simple way? Honestly curious, not provoking, I would really appreciate it.

marcosboger:
@webknjaz I believe in that case, the user should probably leverage [Multiline Markdown content](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/workflow-commands#multiline-markdown-content) capturing the output from the command and `>>` that to `$GITHUB_STEP_SUMMARY`. But that should be done directly by the user configuring this in GHA, so I think no different behavior from pytest-cov needed. What do you think?

beck3905:
I was able to accomplish what I needed using the following composite action that leverages some other community actions:

```
runs:
  using: composite
  steps:
    - name: Run Pytest with Coverage
      shell: bash
      run: |
        poetry run pytest --junitxml=test-report/test-report.xml --cov=. --cov-report=html:coverage/htmlcov --cov-report=xml:coverage/coverage.xml --cov-report=term-missing

    - name: Publish Test Report
      uses: ctrf-io/github-test-reporter@v1
      if: always()
      with:
[...]

PR #712, opened 2025-08-21
Adding Markdown support as a file_choice

marcosboger:
Closes https://github.com/pytest-dev/pytest-cov/issues/688

Option which implements markdown support as a file_choice.

We would diverge a bit from the Coverage.py API, creating our own sort of behavior for the `markdown` option. But might be useful indeed.

Example output:
image

File generated:
image

### DIFF
--- a/docs/reporting.rst
+++ b/docs/reporting.rst
@@ -3,7 +3,7 @@ Reporting
 
 It is possible to generate any combination of the reports for a single test run.
 
-The available reports are terminal (with or without missing line numbers shown), HTML, XML, JSON, LCOV and
+The available reports are terminal (with or without missing line numbers shown), HTML, XML, JSON, MARKDOWN (either in 'write' or 'append' mode to file), LCOV and
 annotated source code.
 
 The terminal report without line numbers (default)::
@@ -49,22 +49,26 @@ The terminal report with skip covered::
 
 You can use ``skip-covered`` with ``term-missing`` as well. e.g. ``--cov-report term-missing:skip-covered``
 
-These four report options output to files without showing anything on the terminal::
+These seven report options output to files without showing anything on the terminal::
 
     pytest --cov-report html
             --cov-report xml
             --cov-report json
+            --cov-report markdown
+            --cov-report markdown-append
             --cov-report lcov
             --cov-report annotate
             --cov=myproj tests/
 
-The output location for each of these reports can be specified. The output location for the XML, JSON and LCOV
+The output location for each of these reports can be specified. The output location for the XML, JSON, MARKDOWN and LCOV
 report is a file. Where as the output location for the HTML and annotated source code reports are
 directories::
 
     pytest --cov-report html:cov_html
             --cov-report xml:cov.xml
             --cov-report json:cov.json
+            --cov-report markdown:cov.md
+            --cov-report markdown-append:cov.md
             --cov-report lcov:cov.info
             --cov-report annotate:cov_annotate
             --cov=myproj tests/
--- a/docs/reporting.rst
+++ b/docs/reporting.rst
@@ -68,7 +68,7 @@ directories::
             --cov-report xml:cov.xml
             --cov-report json:cov.json
             --cov-report markdown:cov.md
-            --cov-report markdown-append:cov.md
+            --cov-report markdown-append:cov-append.md
             --cov-report lcov:cov.info
             --cov-report annotate:cov_annotate
             --cov=myproj tests/
--- a/src/pytest_cov/engine.py
+++ b/src/pytest_cov/engine.py
@@ -255,14 +255,21 @@ class CovController:
                 total = self.cov.json_report(ignore_errors=True, outfile=output)
             stream.write('Coverage JSON written to file %s\n' % (self.cov.config.json_output if output is None else output))
 
-        # Produce markdown report if wanted
+        # Produce Markdown report if wanted.
         if 'markdown' in self.cov_report:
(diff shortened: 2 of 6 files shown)


ionelmc:
This is definitely the way to go (what users are asking anyway). Terminal output is pointless except for debugging. 

We also need to have a `--cov-report=markdown-append:path` option, can you add it?

src/pytest_cov/engine.py
```diff
@@ -255,6 +256,31 @@ def summary(self, stream):
                 total = self.cov.json_report(ignore_errors=True, outfile=output)
             stream.write('Coverage JSON written to file %s\n' % (self.cov.config.json_output if output is None else output))
 
+        # Check that markdown and markdown-append don't point to same file
```
ionelmc, line 259: I would refactor this not have the defaults specified/inferred two times. Also, this ain't the right place to do this check - it's too late - it needs to happen before tests start. Perhaps CovPlugin.__init__ is a better place (cov_report is "prepared" there anyway).

marcosboger: Agreed with both points. I've actually moved the check to `StoreReport` class which is used during argparsing logic. That way, we can throw the error earlier and following the `argparse` style. Now, error is like:

image

What do you think?

ionelmc:
Alright, thank you. I should get around to making a release this weekend I think.

webknjaz (commented):
Thanks for the PR, can't wait to see this released!

I just got back from a vacation and wanted to post a few observations. Some are inline and one more generic question:

Would `--cov-report=markdown:-` work? Using a hyphen is a popular convention for redirecting output to stdout. But its seems like this wouldn't work, right? This might be useful for debugging.
Though, I suppose something like `--cov-report=markdown:/dev/stderr` / `--cov-report=markdown:/dev/stdout` would achieve the same, right?

docs/reporting.rst
```diff
@@ -49,22 +49,26 @@ The terminal report with skip covered::
...
             --cov-report xml
             --cov-report json
+            --cov-report markdown
+            --cov-report markdown-append
             --cov-report lcov
             --cov-report annotate
             --cov=myproj tests/
 
-The output location for each of these reports can be specified. The output location for the XML, JSON and LCOV
+The output location for each of these reports can be specified. The output location for the XML, JSON, MARKDOWN and LCOV
```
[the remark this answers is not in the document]
marcosboger: Yep, will change

docs/reporting.rst
```diff
@@ -49,22 +49,26 @@ The terminal report with skip covered::
...
 You can use ``skip-covered`` with ``term-missing`` as well. e.g. ``--cov-report term-missing:skip-covered``
 
-These four report options output to files without showing anything on the terminal::
+These seven report options output to files without showing anything on the terminal::
 
     pytest --cov-report html
             --cov-report xml
             --cov-report json
+            --cov-report markdown
+            --cov-report markdown-append
```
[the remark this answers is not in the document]
marcosboger: Had the same doubt when making the changes. Do you think adding only one of them would be better? Or maybe already directing one of them to a differen file?

marcosboger:
@webknjaz thanks for the extra insights 👍 

Regarding the question about stdout, I'll test it locally and come back with some results and thoughts.

> Would `--cov-report=markdown:-` work? 

I don't think so too, but maybe this can become a follow-up improvement to allow easier use?

As this was merged already, I'll open a new PR with the doc adjustments and we follow-up there on these points.

docs/reporting.rst
```diff
@@ -49,22 +49,26 @@ The terminal report with skip covered::
 
 You can use ``skip-covered`` with ``term-missing`` as well. e.g. ``--cov-report term-missing:skip-covered``
 
-These four report options output to files without showing anything on the terminal::
+These seven report options output to files without showing anything on the terminal::
```
webknjaz, line 52: It might be a good idea to rephrase this to stop mentioning the exact number, reducing the maintenance burden of having to keep track of this word.

webknjaz:
> I'll open a new PR with the doc adjustments and we follow-up there on these points.

Cool! @ionelmc will still be the one to accept the change as I'm not the maintainer (my `Member` badge shows that I'm in the org but it doesn't reflect the per-repo status).

src/pytest_cov/plugin.py
```diff
@@ -82,6 +82,20 @@ def __call__(self, parser, namespace, values, option_string=None):
         report_type, file = values
         namespace.cov_report[report_type] = file
 
+        # coverage.py doesn't set a default file for markdown output_format
+        if report_type in ['markdown', 'markdown-append'] and file is None:
+            namespace.cov_report[report_type] = 'coverage.md'
+        if all(x in namespace.cov_report for x in ['markdown', 'markdown-append']):
+            self._validate_markdown_dest_files(namespace.cov_report, parser)
```
webknjaz, line 89: It's best to move this into `__init__()` as `__call__()` is happening after `pytest` spends time on running the test and the late error reporting is rather inconvenient. It'd be nicer to tell the user that they need to fix the invocation early.

docs/reporting.rst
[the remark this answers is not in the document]
marcosboger: Adjustes in https://github.com/pytest-dev/pytest-cov/pull/714

Take a look if new format looks better :)

docs/reporting.rst
marcosboger: Adjusted in https://github.com/pytest-dev/pytest-cov/pull/714

PR #718, opened 2025-09-09
Fix syntax for pytest coverage report command

artpelling:
I had trouble getting the advertised command to work. I have changed the docs with fixed syntax.

### DIFF
--- a/docs/reporting.rst
+++ b/docs/reporting.rst
@@ -82,7 +82,7 @@ Example for GitHub Actions with ``markdown-append``:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


ionelmc:
Aaah oops. The trailing dot was causing the problem. The rest will work either way.

webknjaz:
This doesn't look right. I bet it was just the dot that needed to be removed.

artpelling:
To clarify: removing just the dot did not work for me.

ionelmc:
I guess hatch doesn't support the `${foobar}` env var syntax.

tavoletta:
@copilot review

PR #727, opened 2025-11-06
Improve handling of ResourceWarning from sqlite3.

andy-slac:
The patch allows suppressing ResourceWarnings from sqlite3 by specifying filter on command line with -W option.

In our project we have multiple packages with separate `pyproject.toml` file (or even without it). Our build system runs `pytest` on each of the packages, the only way for us to suppress `ResourceWarnings` from sqlite3 is to pass the filter on the command line. This unfortunately does not work, as command line filter is escaped before compiled to regex, and escaped regex does not match non-escaped one that pytest-cov tries to match. I am adding an escaped message to the check so that both variants are checked.

And in general, I do not think it was a good idea for pytest-cov to enable one specific `ResourceWarning` warning. People who care about super-clean code just use `default` filter or `ResourceWarning` with empty message. In our case we do use `-Wd`, but we cannot fix sqlite3 warnings as they come from `sqlalchemy` which has no easy way to close all connections in connection pool (but it is safe to garbage-collect them).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,17 @@
 Changelog
 =========
 
+* Improve handling of ResourceWarning from sqlite3.
+
+  The plugin adds warning filter for sqlite3 ``ResourceWarning`` unclosed database (since 6.2.0).
+  It checks if there is already existing plugin for this message by comparing filter regular expression.
+  When filter is specified on command line the message is escaped and does not match an expected message.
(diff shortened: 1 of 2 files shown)


andy-slac:
Not sure why `tests/check` is failing, can't see any hint in the log. My change is trivial, I don't think it could trigger failure. Could someone check?

ionelmc:
schemastore.org is down, i just need to figure out how or if it's even worth running taplo-lint at all.

PR #741, opened 2026-04-15
Fix RST formatting and double-dash escaping in config

### LINKED ISSUE #743: [Docs] Configuration Page has a Rendering Issue
The "CLI options" listed out in https://pytest-cov.readthedocs.io/en/latest/config.html (from source file [config.rst](https://github.com/pytest-dev/pytest-cov/blob/master/docs/config.rst)) are rendered incorrectly. The first few lines for `--cov`, `--cov-reset`, and `--cov-report` are rendering as a plaintext paragraph rather than as a table with the formatting of the lines beneath. Moreover, a block quote appears in the middle of what should be the explanation of `--cov-report`.

Screenshot of the current rendering below:

Image

OlenaYefymenko:
This patch applies RST formatting to the configuration document, in particular it:

- Adds the `:file:` role for `.coveragerc`, `setup.cfg`, `pyproject.toml`, and `tox.ini`;
- Fixes the CLI options list formatting by switching to `.. option::` directives;
- Removes a stray `>` from the `addopts` reference URL;
- Uses double backticks for `--cov=` and `--cov-report=` to prevent RST from rendering `-` as an em dash.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -22,7 +22,7 @@ CLI options:
 
 .. option:: --cov-config PATH
 
-    Config file for coverage. Default: .coveragerc
+    Config file for coverage. Default: :file:`.coveragerc`
 
 .. option:: --no-cov-on-fail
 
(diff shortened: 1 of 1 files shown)


OlenaYefymenko:
This is how it was rendered before the fix:
Screenshot From 2026-04-16 00-46-17

webknjaz:
This doesn't seem to be enough. And also, the unquoted options are seen in multiple places in that paragraph.

docs/config.rst
```diff
@@ -6,7 +6,7 @@ This plugin provides a clean minimal set of command line options that are added
 further control of coverage use a `coverage config file`_.
 
 CLI options:
```
webknjaz, line 8: I'd try this too, perhaps
```suggestion
CLI options::
```

OlenaYefymenko:
> This doesn't seem to be enough. And also, the unquoted options are seen in multiple places in that paragraph.

Thank you for the review!

OlenaYefymenko:
Checked HTML rendering locally via sphinx-build
Screenshot From 2026-04-24 23-34-48

webknjaz (approved):

webknjaz:
@ionelmc FTR, the linkcheck issues are related to GitHub's recent increase in aggressive rate limiting measures of unauthenticated HTTP requests.

ionelmc:
That is fine. There's no release automation depending on that completing. Thanks everyone.

Issue #728, opened 2025-11-11
Why doesn't pytest-cov fail with a non-zero exit code when code coverage threshold isn't met, when running on multiple directories?


samdewr:
# Summary

## Expected vs actual result
When the average test coverage threshold across multiple modules isn't met, pytest _doesn't_ fail with a non-zero exit code, even though it should. 

# Reproducer

## Versions
OS and hardware
 
```
  Model Name:	MacBook Pro
  Model Identifier:	Mac16,5
  Chip:	                 Apple M4 Max
  System Version:	macOS 15.6 (24G84)
  Kernel Version:	Darwin 24.6.0
```

Pytest versions
```
pytest = "^8.3.4"
pytest-cov = "^6.0.0"
```

## Config
In `pyproject.toml`:

```toml
[tool.coverage.run]
branch = true

[tool.coverage.report]
fail_under = 75
exclude_lines = ['pragma: no cover', 'raise NotImplementedError', '\s*\.\.\.']

[tool.coverage.html]
directory = './coverage'

[tool.coverage.xml]
output = './coverage/cov.xml'
```

## Code
My command: 

```bash
❯ pytest --cov module1 --cov module2 --cov module3 --cov-report --cov-fail-under=75
```

I get the following result: 

```
FAIL Required test coverage of 75.0% not reached. Total coverage: 74.79%
```

But the exit code remains zero:

```bash
❯ echo $?
0
```

As a result, my CI/CD pipeline (triggered upon a PR to `main`) passes just fine even though my coverage threshold isn't met. This causes my `main` branch to have too low test coverage. 

I've tried

- Passing `--cov=module1,module2,module3` instead of having one pytest call with three different `--cov` arguments. This causes the coverage report to fail altogether, because it says that `CoverageWarning: Module module1,module2,module3 was never imported`. I.e., it tries to import `module1,module2,module3` as a single module and it fails, since they are separate modules.
- Passing the to-be-included modules in `pyproject.toml` under `[tool.coverage.run]` in the `include` argument. Problem persists: test coverage threshold failure is noted, but exit code is zero. 
- Interestingly, when you test on each of the modules individually, you do get a non-zero exit code for the module that has too low coverage. 
- On the internet, I can't really seem to find a working solution. Copilot isn't very useful here either.

andywatkins-ancoris:
Thank you @samdewr  for raising this. I also have this issue and found your comment on Stackoverflow, leading to this.

[...]

Issue #638, opened 2024-04-02
"Fail under check" is failing because of ignoring the precession config


tanaydin:
I'm configuring coverage report in pyproject.toml like

```
[tool.coverage.report]
precision = 1
```

and trying to get json report with fail-under configured like

```poetry run pytest --cov --cov-append --cov-fail-under=80.2 --cov-report=json```

but in the following check, it's failing
https://github.com/pytest-dev/pytest-cov/blob/5295ce01c84262cec88f31255e9ac538718f3047/src/pytest_cov/plugin.py#L380

as 

```FAIL Required test coverage of 80.2% not reached. Total coverage: 80.16%```

it should be rounding 80.16 to 80.20 and pass.

PR #751, opened 2026-07-26
Make the fail-under summary agree with the exit code

OnourImpram:
## The defect

`--cov-fail-under` is decided twice, with two different comparisons.

| Where | Predicate |
|---|---|
| `plugin.py` `pytest_sessionfinish` | `should_fail_under(self.cov_total, cov_fail_under, cov_precision)` — coverage.py's own, compares the total **rounded to `[report] precision`** |
| `plugin.py` `pytest_terminal_summary` | `failed = self.cov_total < self.options.cov_fail_under` — the **raw** total |

When the raw total is below the threshold but rounds up to meet it, the two disagree. The run exits `0` and prints a red `FAIL` banner at the same time.

Reproduced on pytest 9.1.1 / pytest-cov 7.1.0 / coverage 7.15.2, with the suite's own `SCRIPT` fixture (88.888…% covered) at the default precision of 0:

```console
$ pytest --cov=... --cov-fail-under=88.9
FAIL Required test coverage of 88.9% not reached. Total coverage: 88.89%
============================= 10 passed in 0.20s ==============================
$ echo $?
0
```

`round(88.888…, 0)` is `89`, so `should_fail_under` says pass and the exit code is `0` — correctly, given the configured precision. Only the banner disagrees.

## The change

One line: the summary uses the predicate the exit-code path already uses.

**The exit code does not change.** Only the banner text and its colour move, so no build that passes today starts failing.

## On the direction of the fix

The inconsistency is not in dispute, but the two open issues want it resolved in *opposite* directions, so this is a maintainer call and I want to be explicit about which way I went.

- **#638** — "'Fail under check' is failing because of ignoring the precession config" — is exactly this: a `FAIL` shown when precision says the threshold was met. This PR fixes it.
- **#728** — the reporter writes that they "rightfully see the message `FAIL …`" and that the problem is the exit code consulting the rounded value. That would be the mirror fix: change `pytest_sessionfinish` to use the raw comparison instead.

I chose to align the banner rather than the exit code because:

1. `should_fail_under` is coverage.py's canonical predicate, and `pytest_sessionfinish` already delegates to it deliberately — the semantic choice has been made in this codebase.
2. `[report] precision` / `--cov-precision` exists to declare the granularity that matters; at precision 0, 88.888…% *is* 89%.
3. It is the non-breaking direction. Switching the exit code to the raw comparison would start failing CI runs that pass today.

If you prefer #728's reading, the change belongs on the other line and I am happy to redo it that way.

## Not changed

The banner still hardcodes `Total coverage: {actual:.2f}%` rather than using `cov_precision`, so at precision 3 it can still print a differently-rounded number than the table above it. Fixing that changes an expected string in `test_cov_min_float_value_not_reached`, so I have left it alone rather than fold an unrelated behaviour change into this PR. Happy to follow up.

## Tests

Two regression tests alongside the existing `test_cov_min_float_value` group — one at default precision, one mirroring #638 with `precision = 1`. Both fail on `master` and pass with the change.

Full suite on Windows / Python 3.13: **205 passed, 6 skipped** (all pre-existing platform skips). `ruff check` and `ruff format --diff` clean.

The remaining fail-under tests are unaffected: their thresholds (88.88 @ p0, 88.89 @ p3, 100 @ p0, 50 @ p0) do not sit in the divergent band, so both predicates return the same boolean.


ionelmc:
Why does this pr show a huge removal/add change on plugin.py? Is that a display bug in github or what's going on?

OnourImpram:
Line endings, not a rewrite. The file had been committed with CRLF, so all 477 lines of `plugin.py` registered as changed. Restored to LF in `f3121a1`; the pull request now shows +43 −1.

The actual change in `plugin.py` is seven lines. `pytest_sessionfinish` decides the exit code with `should_fail_under(cov_total, cov_fail_under, cov_precision)`, while the summary banner compared `cov_total < cov_fail_under` directly. At a precision boundary the two disagree and the run prints that required coverage was not reached while exiting 0, or the reverse. The banner now calls the same predicate, so it cannot contradict the exit status. The remaining diff is one regression test and a changelog entry.

Thanks for looking at it.

PR #759, opened 2026-09-20
Add support for Python 3.15 and drop EOL 3.9

hugovk:
The Python 3.15 release candidate is out. 🚀 

The 3.15 release manager (👋 hi, that's me!) issued a [call to action](https://discuss.python.org/t/python-3-15-0-candidate-2-is-here/?u=hugovk):

> We ***strongly encourage*** maintainers of third-party Python projects to prepare their projects for 3.15 during this phase, and publish Python 3.15 wheels on PyPI to be ready for the final release of 3.15.0, and to help other projects do their own testing. Any binary wheels built against Python 3.15.0 release candidates ***will work*** with future versions of Python 3.15. As always, report any issues to [the Python bug tracker](https://github.com/python/cpython/issues).
> 
> Please keep in mind that this is a preview release and while it's as close to the final release as we can get it, its use is ***not*** recommended for production environments.

Also drop Python 3.9, which was EOL in October 2025. (3.10 will be EOL next month.)

https://devguide.python.org/versions/


hugovk:
Linkcheck and taplo failures are unrelated, will be fixed by https://github.com/pytest-dev/pytest-cov/pull/760 and https://github.com/pytest-dev/pytest-cov/pull/761.

